PR #52, opened 2018-02-20
Text input now supports multiple lines and grows as you type in many …

EtimoOscarNylander:
…lines


victorsmirnov:
@tgkd please have a look.

PR #54, opened 2018-02-20
Added sentry to backend in backoffice as well

EtimoOscarNylander:
Before this can be merged we should look at where to enter the DSN that Sentry needs to work properly - you can approve this ticket but should not merge before we can synchronize this

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -166,6 +166,11 @@
             <artifactId>logstash-logback-encoder</artifactId>
             <version>4.9</version>
         </dependency>
+        <dependency>
+            <groupId>io.sentry</groupId>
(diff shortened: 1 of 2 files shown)


victorsmirnov:
@EtimoOscarNylander are you sure we should have sentry for backend loging? We have kibana for logging now. May be this is Okay?

We can put Sentry settings to the k8s configuration.

EtimoOscarNylander:
@victorsmirnov 
Sentry is less a logging system and more a error tracking system - their use cases are different.
We have sentry in a number of other systems, I have synced this with the rest of the team.

EtimoOscarNylander:
I will merge this now as I need sentry in the testing environment.

PR #121, opened 2018-06-18
fix routing problems

meletios:


src/main/js/app/components/chat/tabs/ClaimsTab.jsx
```diff
@@ -1,26 +1,28 @@
...
-        />
+export default class ClaimsTab extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const { memberClaims, sortClaimsList } = this.props;
+
+    return memberClaims.length ? (
```
hedvigoscar: maybe `memberClaims.length > 0 ? ...`, this could cause some weird bugs in the worst case scenario

src/main/js/app/components/claims/claim-info/ClaimInfo.jsx
```diff
@@ -1,107 +1,116 @@
...
-        claimUpdate(params.id, { state: value }, 'state');
+  constructor(props) {
+    super(props);
+    this.state = {
+      status: "OPEN"
     };
+  }
+
+  getRegistrationDate = date => {
+    return `Registration date: ${moment(date).format("DD MMMM YYYY")} ${moment(
```
hedvigoscar: This moment formatting could be simplified to `moment(date).format('YYYY-MM-DD HH:mm')`

src/main/js/app/components/chat/tabs/ClaimsTab.jsx
meletios: yeap! fixed

src/main/js/app/components/claims/claim-info/ClaimInfo.jsx
meletios: correct! fixed

PR #122, opened 2018-06-27
fix date for claims

meletios:


hedvigoscar:
This should be merged I think!

PR #123, opened 2018-07-10
Feature/change insurance details

meletios:


src/main/java/com/hedvig/backoffice/web/MemberController.java
```diff
@@ -102,12 +104,18 @@ public InsuranceStatusDTO insurance(@PathVariable String memberId, @Authenticati
     }
 
     @GetMapping("/insurance/search")
-    public List<InsuranceStatusDTO> serachInsurance(@RequestParam(name = "state", defaultValue = "", required = false) String state,
+    public List<InsuranceStatusDTO> searchInsurance(@RequestParam(name = "state", defaultValue = "", required = false) String state,
```
hedvigoscar: 👌 👌 👌

src/main/java/com/hedvig/backoffice/web/dto/InsuranceModificationDTO.java
```diff
@@ -0,0 +1,27 @@
+package com.hedvig.backoffice.web.dto;
+
+import java.util.List;
+import java.util.UUID;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
```
hedvigoscar: Could we use `@Value` over `@Data` here? `@Value` is the same as `@Data` but it's also immutable which prevents mistakes later on.

src/main/js/app/components/chat/insurance-list/InsuranceList.jsx
```diff
@@ -0,0 +1,287 @@
...
+        />
+      </React.Fragment>
+    ) : (
+      <Header>No insurances found</Header>
+    );
+  }
+}
+
+InsuranceList.propTypes = {
+  insurance: PropTypes.object.isRequired,
```
hedvigoscar: `PropTypes.object` should generally be avoided as it's a bit too unspecific. No need to change now but it's good to avoid it in the future 🙂

src/main/java/com/hedvig/backoffice/web/dto/InsuranceModificationDTO.java
meletios: done

PR #125, opened 2018-07-27
Feature/monthly payments

meletios:


src/main/java/com/hedvig/backoffice/graphql/ContextBuilder.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import graphql.servlet.GraphQLContext;
+import graphql.servlet.GraphQLContextBuilder;
+
+public class ContextBuilder implements GraphQLContextBuilder {
+
+	@Override
+	public GraphQLContext build(Optional<HttpServletRequest> req, Optional<HttpServletResponse> resp) {
+		// TODO Here we can get the information about who makes the request, so that we can attribute actions made by this person
```
hedvigoscar: We probably need to complete this part before we launch I think.

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.java
```diff
@@ -0,0 +1,44 @@
...
+public class MemberResolver implements GraphQLResolver<Member> {
+    private final PaymentService paymentService;
+	private final DirectDebitStatusLoader directDebitStatusLoader;
+
+	public MemberResolver(PaymentService paymentService, DirectDebitStatusLoader directDebitStatusLoader) {
+        this.paymentService = paymentService;
+        this.directDebitStatusLoader = directDebitStatusLoader;
+    }
+
+    public List<Transaction> getTransactions(Member member) {
```
hedvigoscar: There is an argument to be made to batch load this as well. Lets skip it for now but keep it in mind for the continued development of the payment feature and back-office gql in general

PR #135, opened 2018-09-17
Feature/bordereau

meletios:


src/main/java/com/hedvig/backoffice/services/product_pricing/dto/ProductType.java
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.backoffice.services.product_pricing.dto;
+
+////KNOWTHAT: this enum needs to be the same with the one in product and pricing
```
hedvigoscar: I like this comment! In the future we should look at publishing packages for this so that there's only one source of these.

src/main/java/com/hedvig/backoffice/web/ReportController.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public ReportController(
+    ProductPricingService productPricingService) {
+    this.productPricingService = productPricingService;
+  }
+
+  @GetMapping(value = "/bdx")
+  public ResponseEntity<?> getBDX(HttpServletResponse response, @RequestParam("year") int year,
+    @RequestParam("month") int month,
+    @RequestParam("isStudent") boolean isStudent) {
```
hedvigoscar: `isStudent` is fine for now but will scale poorly. We should make sure to use something like an enum that is in line with IEX for this later on.

src/main/js/app/features/bordereau/index.js
```diff
@@ -0,0 +1,139 @@
...
+    text: "Student",
+    value: "Student"
+  }
+];
+
+export default class Bordereau extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      date: moment().format("YYYY-MM"),
```
hedvigoscar: Maybe we should use `date-fns` instead of Moment here to align ourselves with the rest of our frontend development.

src/main/java/com/hedvig/backoffice/services/product_pricing/dto/ProductType.java
meletios: :smile: yes let do that. i know that Mr Johan has a valid point against that, but lets discuss that with him.

PR #154, opened 2018-11-21
Allow to create new claims

alex-bel:


src/main/java/com/hedvig/backoffice/web/dto/CreateClaimDTO.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.hedvig.backoffice.services.claims.dto.ClaimSource;
+import lombok.Data;
+
+import java.time.LocalDateTime;
+
+@Data
+public class CreateClaimDTO {
+  String memberId;
+  LocalDateTime registrationDate;
```
meletios: It should be Instant this one

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceStub.java
```diff
@@ -243,6 +245,27 @@ public long totalClaims(String token) {
         + stat.getOrDefault(ClaimState.REOPENED.name(), 0L);
   }
 
+  @Override
+  public UUID createClaim(CreateBackofficeClaimDTO claimData, String token) {
+    UUID id = UUID.randomUUID();
+    Claim claim = new Claim();
+    claim.setId("" + id);
+    claim.setDate(claimData.getRegistrationDate().atZone(SWEDEN_TZ).toLocalDateTime());
```
meletios: no need if it was Instant

src/main/java/com/hedvig/backoffice/web/dto/CreateClaimDTO.java
alex-bel: LocalDateTime is correct since we decided to do timezone conversions on backend

PR #155, opened 2018-11-23
Support for 'file_upload' message body type

alex-bel:


src/main/java/com/hedvig/backoffice/services/MessagesFrontendPostprocessor.java
```diff
@@ -0,0 +1,64 @@
...
+  private void processFileUpload(ObjectNode body) {
+    String key = body.path("key").asText();
+    if (StringUtils.isEmpty(key)) {
+      return;
+    }
+
+    URL url = amazonS3.generatePresignedUrl(
+      chatS3Bucket,
+      key,
+      new Date(Instant.now().plus(24, ChronoUnit.HOURS).toEpochMilli()),
```
meletios: it should be valid for 30 minutes

PR #169, opened 2019-01-07
add fire/file claim type support

meletios:


src/main/java/com/hedvig/backoffice/graphql/types/FireDamageClaim.java
```diff
@@ -0,0 +1,49 @@
...
+import com.hedvig.backoffice.services.claims.dto.ClaimData;
+import lombok.Value;
+import lombok.val;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.List;
+
+@Value
+public class FireDamageClaim {
```
hedvigoscar:
```suggestion
@UnionType
public class FireDamageClaim {
```

Also, remember to import it 🙂

meletios: Correcto, gracias :cactus:

PR #182, opened 2019-05-14
Feature/fileUpload end-point

AlexSJ1990:


src/main/java/com/hedvig/backoffice/BackOfficeApplication.java
```diff
@@ -70,6 +71,11 @@ public FilterRegistrationBean secureFilterBean() {
     return filterRegBean;
   }
 
+  @Bean
+  public RestTemplate createRestTemplate() {
```
palmfjord: is this supposed to be here?

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -1,18 +1,24 @@
...
 import com.hedvig.backoffice.graphql.dataloaders.MemberLoader;
 import com.hedvig.backoffice.graphql.types.Claim;
+import com.hedvig.backoffice.graphql.types.FileUpload;
 import com.hedvig.backoffice.graphql.types.Member;
 import com.hedvig.backoffice.graphql.types.MonthlySubscription;
 import com.hedvig.backoffice.services.product_pricing.ProductPricingService;
+import org.springframework.context.annotation.Bean;
 import org.springframework.stereotype.Component;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.client.RestTemplate;
```
palmfjord: no need to commit this file is it? :) Loadsa new imports but no other changes

src/main/java/com/hedvig/backoffice/services/MessagesFrontendPostprocessor.java
```diff
@@ -13,13 +13,16 @@
...
 @Service
 public class MessagesFrontendPostprocessor {
   final AmazonS3 amazonS3;
   final String chatS3Bucket;
 
   public MessagesFrontendPostprocessor(AmazonS3 amazonS3,
-      @Value("${hedvig.chat.s3Bucket}") String chatS3Bucket) {
+      @Value("${hedvig.chat.s3Bucket}") String chatS3Bucket
+  )
+  {
```
palmfjord: very nitty gritty but normally we don't put `{` on a new line after `)`

src/main/java/com/hedvig/backoffice/services/messages/BotServiceStub.java
```diff
@@ -231,4 +222,22 @@ public ExpoPushTokenDTO getExpoPushToken(String memberId, String token) {
...
+  public List<FileUploadDTO> files(String memberId, String token) {
+    List<FileUploadDTO> fileUploads = new ArrayList<>();
+    fileUploads.add(new FileUploadDTO(
+      "ae4e0980-f3ff-11e8-8f52-37917eb5f47f-img_20181112_193429.jpg",
+      Instant.now(),
+      "image.png",
+      "507308289"
+    ));
+    fileUploads.add(new FileUploadDTO(
+      "ae4e0980-f3ff-11e8-8f52-37917eb5f47f-img_20181112_193429.jpg",
```
palmfjord: should we maybe give these files different ids?

src/main/java/com/hedvig/backoffice/services/messages/BotServiceStub.java
```diff
@@ -231,4 +222,22 @@ public ExpoPushTokenDTO getExpoPushToken(String memberId, String token) {
...
   }
+
+  @Override
+  public List<FileUploadDTO> files(String memberId, String token) {
+    List<FileUploadDTO> fileUploads = new ArrayList<>();
+    fileUploads.add(new FileUploadDTO(
+      "ae4e0980-f3ff-11e8-8f52-37917eb5f47f-img_20181112_193429.jpg",
+      Instant.now(),
+      "image.png",
+      "507308289"
```
palmfjord: maybe these member ids should match the member id we're querying? :) so we replace `"507308289"` -> `memberId`

src/main/java/com/hedvig/backoffice/graphql/scalars/URLScalar.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+        return httpUrl.url();
+      }
+    });
+  }
+
+  private static Optional<URL> toURL(Object input) {
+    if (input instanceof URL) {
+      return Optional.of((URL) input);
+    } else if (input instanceof URI) {
```
palmfjord: no need for `else` here as we're always returning in the `if` :) so we could make this slightly cleaner imo by doing something like

```java
if (input instanceof URL) {
  return Optional.of((URL) input);
}

if (input instanceof URI) { 
     try {
        return Optional.of(((URI) input).toURL());
 ...
```

src/main/java/com/hedvig/backoffice/graphql/scalars/URLScalar.java
```diff
@@ -0,0 +1,94 @@
...
+      try {
+        return Optional.of(((URI) input).toURL());
+      } catch (MalformedURLException ignored) {
```
palmfjord: I really think we should return `Optional.empty()` here so we don't accidentally do something more, even if it works like this

palmfjord: that also means we can ditch the last `else`

src/main/java/com/hedvig/backoffice/graphql/scalars/URLScalar.java
```diff
@@ -0,0 +1,94 @@
...
+        throw new CoercingSerializeException(
+          "Expected a 'URL' like object but was '" + input.getClass().getSimpleName() + "'."
+        );
+      }
+
+      @Override
+      public URL parseValue(Object input) throws CoercingParseValueException {
+        String urlStr;
+        if (input instanceof String) {
+          urlStr = String.valueOf(input);
```
palmfjord: maybe we could actually do an early return here? as `parseURL` is very nicely abstracted i don't see why we can't call it in two places. So we get rid of the `else`s and hence the nested ifs

src/main/java/com/hedvig/backoffice/graphql/scalars/URLScalar.java
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+          Optional<URL> url = toURL(input);
+          if (!url.isPresent()) {
```
palmfjord: we could do this more neatly with `return url.orElseThrow(...)` to get rid of some ifs :)

palmfjord:
I'm not really a fan of `if`s and `else`s in case you haven't noticed already hehe 😅 otherwise looks super-nice!

PR #183, opened 2019-05-17
Support for forcing messages to memebers

johantjellden:


meletios:
@johantjellden there is a merge conflict here

PR #184, opened 2019-05-21
call account-service end-point for payments that are scheduled and aw…

AlexSJ1990:
…aiting approval and graphQL query


src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -42,4 +52,28 @@ public GraphQLQuery(ProductPricingService productPricingService, MemberLoader me
...
+  public List<SchedulerStatus> paymentSchedule(ChargeStatus chargeStatus) {
+    List<SchedulerStateDto> schedulerStateDtos = accountService.subscriptionSchedulesAwaitingApproval(chargeStatus);
+
+    if (schedulerStateDtos.isEmpty()) {
+      return new ArrayList<>();
+    }
+
+    List<SchedulerStatus> schedulerStatuses = new ArrayList<>();
+
+    for (SchedulerStateDto schedulerStateDto : schedulerStateDtos) {
```
palmenhq: I believe this could be done with a `.stream().map(...)` to make it sliiightly prettier :)

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -42,4 +52,28 @@ public GraphQLQuery(ProductPricingService productPricingService, MemberLoader me
   public CompletableFuture<Claim> claim(UUID id) {
     return claimLoader.load(id);
   }
+
+  public List<SchedulerStatus> paymentSchedule(ChargeStatus chargeStatus) {
+    List<SchedulerStateDto> schedulerStateDtos = accountService.subscriptionSchedulesAwaitingApproval(chargeStatus);
+
+    if (schedulerStateDtos.isEmpty()) {
```
palmenhq: is this really needed? If `scheduleStateDtos` is empty the for loop will never be triggered and `schedulerStatuses` will be empty when it's returned anyways, no?

src/main/java/com/hedvig/backoffice/graphql/types/SchedulerStatus.java
```diff
@@ -0,0 +1,23 @@
...
+
+import javax.money.MonetaryAmount;
+import java.time.Instant;
+import java.util.Optional;
+import java.util.UUID;
+
+@Value
+public class SchedulerStatus {
+  UUID id;
+  Member member;
```
palmenhq: Hm, I believe GraphQL _should_ be able to resolve this without putting it in this dto. That would make it more efficient because it can utilize the [MemberBatchLoader](https://github.com/HedvigInsurance/back-office/blob/7afa0d90364478a6941cbef75d252a99611c465e/src/main/java/com/hedvig/backoffice/graphql/dataloaders/MemberBatchLoader.java), otherwise we will call member service one call per entry we get here and that's extremely costy

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -56,24 +56,14 @@ public GraphQLQuery(ProductPricingService productPricingService, MemberLoader me
...
-        schedulerStateDto.getChargeStatus(),
-        schedulerStateDto.getChangedBy(),
-        schedulerStateDto.getChangedAt(),
-        schedulerStateDto.getAmount(),
-        schedulerStateDto.getTransactionId()
-      );
-      schedulerStatuses.add(schedulerStatus);
-    }
-    return schedulerStatuses;
+    return schedulerStateDtos.stream().map(schedulerStateDto -> new SchedulerStatus(
```
palmenhq: Awesome! This is details but could we please break it down on separate lines? So it looks something like

```java
schedulerStateDtos
  .stream()
  .map(
    schedulerStateDto -> new SchedulerStatus(
      ...
    )
  )
  .collect(toList())
```

makes it way easier to read what's going on :)

src/main/resources/schema.graphqls
```diff
@@ -264,6 +283,21 @@ enum ClaimPaymentType {
   Automatic
 }
 
+enum ChargeStatus {
+INITIATED,
```
palmenhq: Maybe some indenting here?

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -56,14 +56,18 @@ public GraphQLQuery(ProductPricingService productPricingService, MemberLoader me
...
-      chargeStatus,
-      schedulerStateDto.getChangedBy(),
-      schedulerStateDto.getChangedAt(),
-      schedulerStateDto.getAmount(),
-      schedulerStateDto.getTransactionId()
-    )).collect(Collectors.toList());
+    return schedulerStateDtos
+      .stream()
+      .map
+        (schedulerStateDto -> new SchedulerStatus(
```
palmenhq: this formatting looks slightly off

src/main/java/com/hedvig/backoffice/services/account/dto/SchedulerStateDto.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.NoArgsConstructor;
+
+import javax.money.MonetaryAmount;
+import java.time.Instant;
+import java.util.Optional;
+import java.util.UUID;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Data
```
palmenhq: Could this be a `@Value` instead or no?

PR #187, opened 2019-05-31
Bug fix (memberrs that dont exist in member-service)

vonElfvin:


src/main/resources/schema.graphqls
```diff
@@ -89,7 +89,7 @@ type Account {
 
 type SchedulerState {
   id: ID!
-  member: Member!
+  member: Member
```
hedvigoscar: Lets fix this one on monday ✌️

PR #188, opened 2019-07-02
Auto answer suggestion

hkindbom:


src/main/java/com/hedvig/backoffice/config/ExternalServicesConfig.java
```diff
@@ -60,7 +63,9 @@ public AssetTrackerClient assetTracker(@Value("${tracker.stub:false}") boolean s
   @Bean
   public HopeAutocompleteService hopeAutocompleteService(@Value("${hopeAutocompleteService.stub:false}") boolean stub) {
     AutowireCapableBeanFactory factory = context.getAutowireCapableBeanFactory();
-    return stub ? factory.createBean(HopeAutocompleteServiceStub.class) : factory.createBean(HopeAutocompleteServiceImpl.class);
+    return stub
+      ? factory.createBean(HopeAutocompleteServiceStub.class)
+      : factory.createBean(HopeAutocompleteServiceImpl.class); //: factory.createBean(HopeAutocompleteServiceStub.class);
```
palmenhq: commented out code 🙈

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
```diff
@@ -48,16 +47,18 @@
   private final ClaimLoader claimLoader;
   private final ClaimsService claimsService;
   private final AccountService accountService;
+  private AutoAnswerSuggestionService autoAnswerSuggestionService;
```
palmenhq: final plz :)

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
```diff
@@ -68,6 +69,11 @@ public GraphQLMutation(PaymentService paymentService, PersonnelService personnel
     return memberLoader.load(id);
   }
 
+  public AutoLabelDTO autoLabelQuestion(String question, String label, String memberId,  List<String> messageId) {
```
palmenhq: `messageId` is a a list right? So it's multiple message ids = name should be `messageIds`

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
```diff
@@ -68,6 +69,11 @@ public GraphQLMutation(PaymentService paymentService, PersonnelService personnel
...
+    autoAnswerSuggestionService.autoLabelQuestion(question, label, memberId, messageId);
+    return new AutoLabelDTO("test reply auto labeling");
```
palmenhq: 🙈

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -45,6 +48,10 @@ public GraphQLQuery(ProductPricingService productPricingService, MemberLoader me
         .collect(Collectors.toList());
   }
 
+  public SuggestionDTO getAnswerSuggestion(String question) {
+    return this.autoAnswerSuggestionService.getAnswerSuggestion(question);
```
palmenhq: doesn't matter too much i guess but no need for `this` in this context

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/AutoAnswerSuggestionServiceStub.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+
+
+@Slf4j
+public class AutoAnswerSuggestionServiceStub implements AutoAnswerSuggestionService {
+
+  @Override
+  public SuggestionDTO getAnswerSuggestion(String question) {
+
+    String sampleAnswer = "[\n" +
```
palmenhq: this is waaaaay to much multiline text, it's will be virtually impossible to maintain. We should convert this to DTOs. If you insist on keeping this as a blob of text (which I **HIGHLY** oppose) you could make this into DTOs and use an `ObjectMapper` to convert it to text

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/AutoAnswerSuggestionServiceStub.java
```diff
@@ -0,0 +1,70 @@
...
+      "        \"text\": \"hej\"\n" +
+      "    }\n" +
+      "]";
+
+    return new SuggestionDTO(sampleAnswer);
+  }
+
+  @Override
+  public void autoLabelQuestion(String question, String label, String memberId,  List<String> messageId) {
+    System.out.println(question);
```
palmenhq: Use a logger please :)

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/SuggestionDTO/AutoLabelDTO.java
```diff
@@ -0,0 +1,17 @@
+package com.hedvig.backoffice.services.autoAnswerSuggestion.SuggestionDTO;
+
+import lombok.Value;
+
+
+@Value
+public class AutoLabelDTO {
+  String message;
+
+  public AutoLabelDTO(String answer) {
```
palmenhq: No need to make a constructor when you use `@Value`

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/SuggestionDTO/AutoLabelDTO.java
```diff
@@ -0,0 +1,17 @@
+package com.hedvig.backoffice.services.autoAnswerSuggestion.SuggestionDTO;
+
+import lombok.Value;
+
+
+@Value
+public class AutoLabelDTO {
+  String message;
```
palmenhq: `private String`*

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/SuggestionDTO/SuggestionDTO.java
```diff
@@ -0,0 +1,58 @@
+package com.hedvig.backoffice.services.autoAnswerSuggestion.SuggestionDTO;
```
palmenhq: i don't like this package, why do we need to put the `SuggestonDTO` class inside a `SuggestionDTO` package? Package inception 😬 if we put multiple DTO's (which it seems like we're doing) here then we should rename this package to for example just `dtos` or so

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/SuggestionDTO/SuggestionDTO.java
```diff
@@ -0,0 +1,58 @@
...
+
+  String intent;
+  String reply;
+  String text;
+  ArrayList<ReplyEntry> allRepliesArray;
+
+  public SuggestionDTO(String autoAnswerServiceReply) {
+
+    //strip [ and ] in beginning and end of string
+    String result = autoAnswerServiceReply.substring(1, autoAnswerServiceReply.length() - 2);
```
palmenhq: We should use a proper JSON parser, this is WAY too shaky. We shouldn't fail for example if you start/end the input with newline or space - which would happen in this case

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/SuggestionDTO/SuggestionDTO.java
```diff
@@ -0,0 +1,58 @@
...
+
+    allRepliesArray = new ArrayList<>();
+
+    try {
+      JSONObject obj = new JSONObject(result);
```
palmenhq: please make proper DTO's and use jackson's object mapper, AFAIK this class is made for configurations, not for use in request/response objects

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/SuggestionDTO/SuggestionDTO.java
```diff
@@ -0,0 +1,58 @@
...
+
+    } catch (JSONException e) {
+
+      e.printStackTrace();
+    }
+
+  }
+
+  public class ReplyEntry {
+    String intent;
```
palmenhq: `private` + getters. or `@Value` plz :)

src/main/java/com/hedvig/backoffice/services/hopeAutocomplete/HopeAutocompleteServiceStub.java
```diff
@@ -1,12 +1,21 @@
...
 import java.util.ArrayList;
 import java.util.List;
 
+import static java.util.Objects.requireNonNull;
+
 public class HopeAutocompleteServiceStub implements HopeAutocompleteService {
+  private final HopeAutocompleteClient client;
+
+  public HopeAutocompleteServiceStub(final HopeAutocompleteClient client) {
+    this.client = requireNonNull(client);
```
palmenhq: love it! really nice

src/main/java/com/hedvig/backoffice/services/hopeAutocomplete/HopeAutocompleteServiceStub.java
```diff
@@ -17,6 +26,8 @@
 
   @Override
   public void selectSuggestion(SelectSuggestionRequestBodyDto requestBody) {
-    // noop
+
+    //client.selectAutocompleteSuggestion(RemoteAutocompleteSuggestionRequestBodyDto.from(requestBody));
```
palmenhq: woops 😬

src/main/resources/application.yaml
```diff
@@ -59,6 +59,10 @@ tracker:
   stub: false
 
 hopeAutocompleteService:
+  baseUrl: http://localhost:5000
```
palmenhq: maybe they shouldn't both run on port `5000` though

src/main/resources/schema.graphqls
```diff
@@ -22,6 +24,22 @@ type MutationType {
   setClaimInformation(id: ID!, information: ClaimInformationInput!): Claim
   updateReserve(id: ID!, amount: MonetaryAmount!): Claim
   setCoveringEmployee(id: ID!, coveringEmployee: Boolean!): Claim
+  autoLabelQuestion(question: String!, label: String!, memberId: String, messageId: [String!]): AutoLabel
+}
+
+type Suggestion {
+  intent: String
```
palmenhq: without suffixing the types with `!` they're all considered nullable. I'm assuming all of this is not nullable. This should probably be `String!` for example

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/SuggestionDTO/SuggestionDTO.java
```diff
@@ -0,0 +1,58 @@
...
+        if (allReplies.get(key) instanceof JSONObject) {
+
+          JSONObject replyObj = allReplies.getJSONObject(key);
+          allRepliesArray.add(new ReplyEntry(key, replyObj.getString("reply")));
+        }
+      }
+
+    } catch (JSONException e) {
+
+      e.printStackTrace();
```
palmenhq: we should propagate exceptions or log them. For example if this happens we will probably receive a `NullPointerException` later, because it won't put the data on this class' properties and we will try to use them later, and we will have little to no idea why they are null. also always use a logger instead of printing to stdout

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/AutoAnswerSuggestionServiceImpl.java
```diff
@@ -0,0 +1,36 @@
+package com.hedvig.backoffice.services.autoAnswerSuggestion;
+
+import com.hedvig.backoffice.services.autoAnswerSuggestion.SuggestionDTO.SuggestionDTO;
+import com.hedvig.backoffice.services.autoAnswerSuggestion.SuggestionDTO.AutoLabelDTO;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.List;
+
+
+@Slf4j
```
palmenhq: we don't need this if we're not logging anything, which we're not

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
hkindbom: I know but I think I must return something in the mutation

palmenhq: yes that is true, but you could for example change it to a `Boolean` and return true always :) that's what we've done in other places

hkindbom: Got it!

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/AutoAnswerSuggestionServiceStub.java
hkindbom: I just use this while developing to mock the request from python. Is it fine?

palmenhq: I mean i don't really see why. It would imo be way easier to just construct the SuggestionDTO without parsing n stuff.

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/AutoAnswerSuggestionServiceClient.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import java.util.List;
+
+@FeignClient(
+  name = "autoAnswerSuggestion-service",
+  url = "${autoAnswerSuggestionService.baseUrl}",
```
palmenhq: i think this is what is failing the build, it needs to go into `application(-test)?.yml` so spring knows what to do with it, alternatively that you put a default value in this interpolation

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/DTOs/AutoLabelDTO.java
```diff
@@ -0,0 +1,13 @@
+package com.hedvig.backoffice.services.autoAnswerSuggestion.DTOs;
```
palmenhq: For future reference, try to keep package names lowercase as long as you can, as that's kind of a convention in java

palmenhq: But as git is case insensitive it's a bit of a hazzle to rename these directories to lowercase, so if you want to leave it imo its fine

src/main/java/com/hedvig/backoffice/services/autoAnswerSuggestion/DTOs/SuggestionDTO.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.ArrayList;
+
+@Value
+public class SuggestionDTO {
+
+  String intent;
+  String reply;
+  String text;
+  Float confidence;
```
palmenhq: actually i believe this should be a `double`, as both python and graphql `Float`s have double precision afaik

hkindbom: Fixed it!

PR #190, opened 2019-07-12
Pre-signed URLs for Claim recordings

hedvigoscar:


src/main/java/com/hedvig/backoffice/graphql/types/Claim.java
```diff
@@ -1,7 +1,10 @@
 package com.hedvig.backoffice.graphql.types;
 
+import com.amazonaws.services.s3.AmazonS3;
 import com.hedvig.backoffice.services.claims.dto.ClaimData;
 import lombok.Value;
+import lombok.val;
```
meletios: same here

src/main/java/com/hedvig/backoffice/graphql/types/Claim.java
```diff
@@ -1,7 +1,10 @@
 package com.hedvig.backoffice.graphql.types;
 
+import com.amazonaws.services.s3.AmazonS3;
```
meletios: you can discard the imports on this file.It seems that are not needed

src/main/java/com/hedvig/backoffice/graphql/types/Claim.java
hedvigoscar:
```suggestion
```

src/main/java/com/hedvig/backoffice/graphql/types/Claim.java
hedvigoscar:
```suggestion
```

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceImpl.java
```diff
@@ -121,7 +131,18 @@ public long totalClaims(String token) {
 
   @Override
   public List<Claim> getClaimsByIds(List<UUID> ids) {
-    return client.getClaimsByIds(new ClaimsByIdsDto(ids));
+    return client.getClaimsByIds(new ClaimsByIdsDto(ids))
+      .stream()
+      .map(c -> {
+        val signedAudioUrl = signAudioUrl(c.getAudioURL());
+        val claimWithSigned = c.toBuilder().audioURL(signedAudioUrl).build();
```
meletios: Shall we assign the rest of the properties in the builder also?

hedvigoscar: I'm not actually sure the builder generates the inherited properties. I think that's why the properties were not automatically assigned by `toBuilder()`

meletios: oh yes, it extends ClaimBackOffice 💥  👍

PR #193, opened 2019-07-29
Add correct account entry types to back-office

vonElfvin:


src/main/java/com/hedvig/backoffice/services/account/dto/AccountEntryType.java
```diff
@@ -6,5 +6,6 @@
   CAMPAIGN,
   PAYOUT,
   CHARGE,
-  RESERVED
```
palmenhq: is this intentional?

vonElfvin: Yes, we don't use reserved in account-service.

PR #198, opened 2019-08-05
Questions to tickets

fwald:
Made integration between Question-service and ticket-service

### DIFF
--- a/src/main/java/com/hedvig/backoffice/services/tickets/TicketServiceImpl.java
+++ b/src/main/java/com/hedvig/backoffice/services/tickets/TicketServiceImpl.java
@@ -1,6 +1,8 @@
 package com.hedvig.backoffice.services.tickets;
 
 import com.hedvig.backoffice.graphql.types.RemindNotification;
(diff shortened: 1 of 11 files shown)


src/main/java/com/hedvig/backoffice/services/tickets/TicketServiceImpl.java
```diff
@@ -53,4 +55,37 @@ public void changePriority(UUID ticketId, float newPriority, String modifiedBy)
...
   }
 
+  @Override
+  public void createTicketFromQuestions(QuestionGroupDTO questionGroup) {
+    List<QuestionDTO> questions = questionGroup.getQuestions();
+    if (questions.isEmpty()) {
+      return;
+    }
+
+    int lastQuestionIndex = 0;
```
AlexSJ1990: could we make this simpler by doing something like 

```
    QuestionDTO lastQuestion = questions.stream()
      .sorted(Comparator.comparing(QuestionDTO::getDate).reversed())
      .collect(Collectors.toList()).get(0);
```

PR #199, opened 2019-08-07
Questions to ticket impl

fwald:
In this pull request we add the feature to resolve a ticket connected to a question when we click the "done" button.
Also, when we create a new ticket from a question we will send it to a separate endpoint. This is done so that we can see if there already exists a ticket associated with that question-group later on, in ticket-service.

Also added end-points and updated a gql query for when we get all tickets from ticket-service. Now we can  request to only get tickets that have/have not been resolved. This is to ease the burden of the front-end in the future, not asking for more tickets than necessary.

### DIFF
--- a/src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
+++ b/src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
@@ -2,6 +2,7 @@ package com.hedvig.backoffice.graphql;
 
 import com.coxautodev.graphql.tools.GraphQLMutationResolver;
 import com.google.common.collect.ImmutableMap;
+import com.hedvig.backoffice.domain.Personnel;
 import com.hedvig.backoffice.graphql.dataloaders.ClaimLoader;
 import com.hedvig.backoffice.graphql.dataloaders.MemberLoader;
 import com.hedvig.backoffice.graphql.types.Claim;
@@ -17,6 +18,8 @@ import com.hedvig.backoffice.services.claims.dto.ClaimPaymentType;
 import com.hedvig.backoffice.services.claims.dto.*;
 import com.hedvig.backoffice.services.payments.PaymentService;
 import com.hedvig.backoffice.services.personnel.PersonnelService;
+import com.hedvig.backoffice.services.questions.QuestionNotFoundException;
+import com.hedvig.backoffice.services.questions.QuestionService;
 import com.hedvig.backoffice.services.tickets.TicketService;
 import com.hedvig.backoffice.services.tickets.dto.*;
 import com.hedvig.backoffice.services.tickets.dto.TicketStatus;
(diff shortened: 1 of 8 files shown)


src/main/java/com/hedvig/backoffice/services/tickets/TicketServiceImpl.java
```diff
@@ -18,8 +18,13 @@ public TicketServiceImpl(TicketServiceClient ticketServiceClient) {
   }
 
   @Override
-  public List<TicketDto> getAllTickets() {
-    return this.ticketServiceClient.getTickets();
+  public List<TicketDto> getAllTickets(Boolean onlyResolvedTickets) {
+    if (onlyResolvedTickets == null) {
```
vonElfvin: I would rather send an enum than a boolean which can be null. Rather change it to "getAllTicketsOfStatus" and send the status and default to all

src/main/resources/schema.graphqls
```diff
@@ -12,7 +12,7 @@ type QueryType {
   items(payload: Payload!): ItemSearch!
   prices(date: String!, ids: [String!]!): [PricePoint!]!
   ticket (id: ID!): Ticket
-  tickets: [Ticket!]!
+  tickets (resolved: Boolean): [Ticket!]!
```
vonElfvin:
```suggestion
  tickets (status: TicketStatus): [Ticket!]!
```

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
```diff
@@ -52,11 +57,13 @@
...
   private final TicketService ticketService;
   private final AutoAnswerSuggestionService autoAnswerSuggestionService;
-
+  private final QuestionService questionsService;
 
   public GraphQLMutation(PaymentService paymentService, PersonnelService personnelService,
                          MemberLoader memberLoader, ClaimLoader claimLoader, ClaimsService claimsService,
-                         AccountService accountService, TicketService ticketService, AutoAnswerSuggestionService autoAnswerSuggestionService) {
+                         AccountService accountService, TicketService ticketService, AutoAnswerSuggestionService autoAnswerSuggestionService,
+                        QuestionService questionsService
```
vonElfvin: separate lines pls
```suggestion
  AccountService accountService,
  TicketService ticketService,
  AutoAnswerSuggestionService autoAnswerSuggestionService,
  QuestionService questionsService
```

src/main/java/com/hedvig/backoffice/services/tickets/TicketServiceClient.java
```diff
@@ -23,9 +23,18 @@
   @GetMapping("/_/tickets/all")
   List<TicketDto> getTickets();
 
+  @GetMapping("/_/tickets/all/unresolved")
+  List<TicketDto> getUnresolvedTickets();
```
vonElfvin:
```suggestion
  List<TicketDto> getTicketsOfStatus(TicketStatus status);
```

PR #201, opened 2019-08-09
Ticket history feature

fwald:
Introduces support for new ticket history feature in Hedgehog. New DTOs have been created as well as a new graphQL query.


src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -116,6 +117,11 @@ public TicketDto ticket(UUID  id) {
     return this.ticketService.getTicketById(id);
   }
 
+  public FullHistoryTicketDto ticketWithFullHistory(UUID  id) {
```
vonElfvin:
```suggestion
  public FullHistoryTicketDto getFullTicketHistory(UUID  id) {
```

PR #204, opened 2019-08-21
debt service integration - provide person data to hedgehog

AlexSJ1990:


src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.java
```diff
@@ -92,4 +98,19 @@ public SanctionStatus getSanctionStatus(Member member) {
     }
     return fileUploads;
   }
-}
+
+  public Person getPerson(Member member) {
+    String memberId = member.getMemberId();
+    Optional<PersonDTO> personDTO = memberService.getPerson(memberId);
+
+    personDTO.orElseThrow(() -> new NullPointerException("No person exists!"));
```
palmenhq: Maybe we could make the message something like `"No person with id " + memberId + " + existst!"` just to make it easier to backtrack the cause if this ever happens

PR #205, opened 2019-08-21
Kill the home controller 🔪

palmenhq:
It's absolutely crazy how much logging errors we're getting from this controller because we removed the template it returns


palmenhq:
@vonElfvin i think we're having some trouble with BO in staging so i'm just gonna wait until tomorrow before I merge so i can figure that out then, i have somewhere i need to bee soon 😬

PR #215, opened 2019-10-22
Change some dto variable names so the integration is intact

vonElfvin:


src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -77,15 +78,17 @@ class MemberResolver(
...
-    )
+    return try {
+      val personDTO = memberService.getPerson(memberId)
+      Person(
+        personFlags = listOf(personDTO.flags.debtFlag),
+        debt = personDTO.debt,
+        whitelisted = personDTO.whitelisted,
+        status = personDTO.status
+      )
+    } catch (exception: Exception) {
```
palmenhq: I think we should be more granular than `Exception`, this is scary since we might say "No person found" when really it was an `InvocationTargetException` or something

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -77,15 +78,17 @@ class MemberResolver(
...
+      throw NullPointerException("No person found (exception: $exception)")
```
palmenhq: We should pass the exception as a `cause` rather than as a string, makes the stack traces way easier to read

PR #218, opened 2019-11-12
Quotes

palmenhq:
dependent on HedvigInsurance/product-pricing#361 and HedvigInsurance/underwriter#22

pinig @hedvigoscar and @FredrikAreschoug regarding the gql schema


src/main/java/com/hedvig/backoffice/graphql/resolvers/QuoteResolver.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.coxautodev.graphql.tools.GraphQLResolver
+import com.hedvig.backoffice.graphql.dataloaders.MemberLoader
+import com.hedvig.backoffice.graphql.types.Member
+import com.hedvig.backoffice.graphql.types.Quote
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteDto
+import org.springframework.stereotype.Component
+import java.util.concurrent.CompletableFuture
+
+@Component
+class QuoteResolver(private val memberLoader: MemberLoader) : GraphQLResolver<Quote> {
```
palmenhq: i think this is not needed now

src/main/java/com/hedvig/backoffice/services/underwriter/UnderwriterService.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.hedvig.backoffice.graphql.types.QuoteInput
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteDto
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteInputDto
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteResponseDto
+import com.hedvig.backoffice.web.dto.CreateQuoteFromProductDto
+import java.time.LocalDate
+import java.util.UUID
+
+interface UnderwriterService {
```
palmenhq: shall we make a stub also perhaps?

src/main/java/com/hedvig/backoffice/services/underwriter/UnderwriterServiceImpl.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteDto
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteInputDto
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteRequestDto
+import com.hedvig.backoffice.services.underwriter.dtos.QuoteResponseDto
+import com.hedvig.backoffice.web.dto.CreateQuoteFromProductDto
+import org.springframework.stereotype.Component
+import java.time.LocalDate
+import java.util.UUID
+
+@Component // TODO make proper impl
```
palmenhq: ??

src/main/java/com/hedvig/backoffice/services/underwriter/dtos/QuoteRequestDto.kt
```diff
@@ -0,0 +1,85 @@
+package com.hedvig.backoffice.services.underwriter.dtos
+
+import com.hedvig.backoffice.services.product_pricing.dto.ExtraBuildingDTO
+import com.hedvig.backoffice.services.product_pricing.dto.ExtraBuildingType
+import com.hedvig.backoffice.services.product_pricing.dto.ProductType
+import java.time.LocalDate
+import java.util.UUID
+
+data class QuoteRequestDto(
```
palmenhq: one of these dtos have to be redundant 😂

palmenhq: after looking, not really

src/main/resources/schema.graphqls
```diff
@@ -140,8 +144,168 @@ type Member {
...
+type ApartmentQuoteData {
+  id: ID!
+  ssn:  String
+  firstName: String
+  lastName: String
+  street: String
+  city: String
+  zipCode: String
+  householdSize: Int
+  livingSpace: Int
```
FredrikAreschoug: Could create an interface for these

src/main/resources/schema.graphqls
```diff
@@ -140,8 +144,168 @@ type Member {
...
+  numberOfBathrooms: Int
+  extraBuildings: [ExtraBuildingInput!]!
+  isSubleted: Boolean
+}
+
+input ExtraBuildingInput {
+  type: String!
+  area: Int!
+  hasWaterConnected: Boolean!
+  displayName: String
```
FredrikAreschoug: Maybe we should remove this since we do not support edit of the display name currently. But I would love to add that feature :)

src/main/java/com/hedvig/backoffice/services/underwriter/dtos/QuoteDto.kt
```diff
@@ -0,0 +1,152 @@
...
+    val numberOfBathrooms: Int? = null,
+    val extraBuildings: List<ExtraBuilding>? = null,
+    val isSubleted: Boolean? = null,
+    val floor: Int = 0
+  ) : QuoteData() {
+    companion object {
+      fun from(houseQuoteDataInput: HouseQuoteDataInput): HouseQuoteData =
+        HouseQuoteData(
+          id = null,
+          ssn = houseQuoteDataInput.firstName,
```
FredrikAreschoug: this does not look right

src/main/java/com/hedvig/backoffice/graphql/types/Quote.kt
```diff
@@ -0,0 +1,155 @@
...
+      Quote(
+        id = quote.id,
+        currentInsurer = quote.currentInsurer,
+        memberId = quote.memberId,
+        startDate = quote.startDate,
+        price = quote.price,
+        data = when {
+          quote.data is com.hedvig.backoffice.services.underwriter.dtos.QuoteData.ApartmentQuoteData -> QuoteData.ApartmentQuoteData(
+            id = quote.data.id!!,
+            ssn = quote.data.firstName,
```
FredrikAreschoug: this does not look right

src/main/java/com/hedvig/backoffice/graphql/types/Quote.kt
```diff
@@ -0,0 +1,155 @@
...
+            street = quote.data.street,
+            zipCode = quote.data.zipCode,
+            city = quote.data.city,
+            livingSpace = quote.data.livingSpace,
+            householdSize = quote.data.householdSize,
+            subType = quote.data.subType
+          )
+          quote.data is com.hedvig.backoffice.services.underwriter.dtos.QuoteData.HouseQuoteData -> QuoteData.HouseQuoteData(
+            id = quote.data.id!!,
+            ssn = quote.data.firstName,
```
FredrikAreschoug: same here

src/main/resources/schema.graphqls
```diff
@@ -37,6 +37,10 @@ type MutationType {
   questionIsDone(memberId: ID!): Boolean
   whitelistMember(memberId: ID!): Boolean
   backfillSubscriptions(memberId: ID!): Member!
+  activateQuote(id: ID!, activationDate: LocalDate!, terminationDate: LocalDate): Quote!
+  # Creates a quote from a product and returns the quote id
+  createQuoteFromProduct(memberId: ID!, quoteData: QuoteFromProductInput!): ID!
```
hedvigoscar: This could probably return a proper Quote, right?

src/main/resources/schema.graphqls
palmenhq: oh. Hm. For what reason don't we support it?

src/main/resources/schema.graphqls
```diff
@@ -569,5 +733,7 @@ scalar LocalDate
 scalar LocalTime
 # A String-representation of `java.time.LocalDateTIme`. ex: `"2018-06-11T20:08:30.123456"`
 scalar LocalDateTime
+# A String-representation of `java.time.Instant`. ex: `"2018-06-11T20:08:30.123456"`
```
hedvigoscar: Is this how Instants are represented in String-format? I thought they were Epoch numbers 😮

src/main/resources/schema.graphqls
```diff
@@ -140,8 +144,168 @@ type Member {
...
+
+  ancillaryArea: Int
+  yearOfConstruction: Int
+  numberOfBathrooms: Int
+  extraBuildings: [ExtraBuilding!]!
+  isSubleted: Boolean
+}
+
+type ExtraBuilding {
+  type: String!
```
hedvigoscar: Is this not represented by an enum in Underwriter? I think enums would be preferable here as well

src/main/resources/schema.graphqls
palmenhq: 🤷‍♀️ its configurable by jackson. But then we need to change [the existing scalar converter](https://github.com/HedvigInsurance/back-office/blob/9faad8b42adb685f6973bcc7c24bc12c50d97698/src/main/java/com/hedvig/backoffice/graphql/scalars/InstantScalar.java) and it just seems tedious and im lazy

src/main/resources/schema.graphqls
palmenhq: it is, its just very long and i took the easy path 😅 i can change it

src/main/resources/schema.graphqls
palmenhq: yes, but i don't really see the reason. We'll need to do a "proper" refetch query either way

hedvigoscar: Usually you can write the item fetched from the 'post-mutation query' to the cache if desirable. Additionally I think it's not a bad idea from an API flexibility standpoint 🙂

Under such a scheme, if you're only interested in the ID you can easily query it out from the resulting `Quote`-object. Plus this code can reuse the same data-resolving facilities created for the `Quote`-object 🙂

src/main/resources/schema.graphqls
palmenhq: coming back to it - no its strings in the underwriter also it _is_ however an enum in product-pricing. Idk if its worth the hassle though? Both now and maintaining the same enum in at least 4 services? (underwriter, product-pricing, back-office, hedgehog (+ app, web onboarding, giraffe ...))?

hedvigoscar: I think Enums are far easier to reason about - I'd vote for using enums 🙂

palmenhq: i think this makes an exceptionally good case for centralizing the inter-service comms like we've discussed. I say lets leave it as strings for now and move it to be proper enums once we have a more sustainable solution later

src/main/resources/schema.graphqls
FredrikAreschoug: This is how we fetch the display name https://github.com/HedvigInsurance/product-pricing/blob/c833920b1e4768dfd90c6e90338791509ee6ebf4/src/main/java/com/hedvig/productPricing/service/web/dto/ExtraBuildingDTO.kt#L32 . So as an input I don't see how this would work

palmenhq: aha interesting. Then i will change 😬👍

palmenhq:
@johantjellden that is true, also in case we need to rollback that'd be much easier to do

src/main/resources/schema.graphqls
hedvigoscar: I feel like we're going to gain very little by using strings here over Enums to be honest 😞

palmenhq: i feel the reverse 😂 any third opinions?

hedvigoscar: Paging @FredrikAreschoug @johantjellden @sampettersson @meletios

vonElfvin: I like strings when it's a high probability the enum will change in the future so you don't have to change it everywhere. For this case I think the type is unlikely to change and thus I prefer enum. Just my 2 cents.

src/main/java/com/hedvig/backoffice/graphql/types/Quote.kt
```diff
@@ -0,0 +1,155 @@
...
+data class Quote(
+  val id: UUID,
+  val createdAt: Instant,
+  val price: BigDecimal? = null,
+  val productType: ProductType,
+  val state: QuoteState,
+  val initiatedFrom: String,
+  val attributedTo: String,
+  val data: QuoteData,
+  val currentInsurer: String? = null,
```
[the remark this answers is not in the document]
sampettersson: @vonElfvin hehe I actually renamed it to `previousInsurer` in giraffe this summer 🕵 So I agree that it's better naming! https://github.com/HedvigInsurance/giraffe/pull/465

src/main/resources/schema.graphqls
palmenhq: Hm, coming back to this again i kind of agree, feels silly to switch data types across the stack. Fixing the tediousness is out of scope for this task i believe. Ill switch to enums 👍

src/main/resources/schema.graphqls
```diff
@@ -140,8 +144,179 @@ type Member {
...
+}
+
+enum ApartmentSubType {
+  BRF,
+  RENT,
+  SUBLET_RENTAL,
+  SUBLET_BRF,
+  STUDENT_BRF,
+  STUDENT_RENT,
+  HOUSE
```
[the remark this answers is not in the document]
johantjellden: Yes this should probably be removed.

vonElfvin:
Just to make sure. And this may be cross service, but I'm quite confident we don't need:
1. Number of floors
2. Safety increasers
3. "isInsuredAtOtherCompany" as a boolean
4. goodToHaveItems
When creating products, just make sure they are not present now that we are making things better+newer.

Also, the discussion of wether perils should be included in the creation of a product/contract can be discussed.

palmenhq:
@vonElfvin 
amaze input!

1, 2, 3. Makes sense. will fix.
4. What is `goodToHaveItems`?
Perils: interesting, but to not creep the scope too much i will ignore this for now. We _should_ however look into it in for example HedvigInsurance/product-pricing#362 or its corresponding master branch

PR #221, opened 2019-11-25
Complete quote on update and improve logging

palmenhq:
Causing some issues where the quote remains incomplete if the quote fails to complete initially


src/main/java/com/hedvig/backoffice/services/underwriter/UnderwriterServiceImpl.kt
```diff
@@ -10,15 +10,21 @@ import com.hedvig.backoffice.services.underwriter.dtos.QuoteDto
...
 
+private val logger: Logger = getLogger(UnderwriterServiceImpl::class.java)
+
 class UnderwriterServiceImpl(
   private val underwriterClient: UnderwriterClient,
   private val memberService: MemberService
 ) : UnderwriterService {
   override fun createAndCompleteQuote(memberId: String, quoteDto: CreateQuoteFromProductDto): QuoteResponseDto {
     val member = memberService.findByMemberId(memberId, "")
+    logger.info("Creating quote for member $memberService")
```
meletios: do you mean `$memberId` ?

palmenhq: woops, nice catch. i'll push a fix directly on master if that's ok

PR #225, opened 2019-11-26
Revert "Revert "Feature/integrate with claims service add claims files""

AlexSJ1990:
Reverts HedvigInsurance/back-office#223

### DIFF
--- a/src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceStub.java
+++ b/src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceStub.java
@@ -8,6 +8,7 @@ import com.hedvig.backoffice.services.members.MemberService;
 import com.hedvig.backoffice.services.settings.SystemSettingsService;
 import java.time.Instant;
 import com.hedvig.backoffice.web.dto.MemberStatus;
+import java.util.stream.Stream;
 import lombok.val;
 import org.apache.commons.lang3.RandomUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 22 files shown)


src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceStub.java
```diff
@@ -277,6 +286,43 @@ public void markEmployeeClaim(EmployeeClaimRequestDTO dto, String token) {
...
 
+  @Override
+  public ResponseEntity<Void> uploadClaimsFiles(
+    String claimId,
+    MultipartFile[] claimFiles,
+    String memberId
+  ) throws IOException {
+    ArrayList claimFileDtos = new ArrayList<ClaimFileDTO>();
+
+    for (MultipartFile claimFile : claimFiles) {
```
palmenhq: didn't we fix this for loop?

src/main/java/com/hedvig/backoffice/services/claims/UploadClaimFiles.kt
```diff
@@ -0,0 +1,47 @@
...
+    contentType: String,
+    data: ByteArray,
+    claimId: String,
+    fileName: String,
+    memberId: String): UploadResult {
+    val metadata = ObjectMetadata()
+    metadata.contentType = contentType
+    metadata.contentLength = data.size.toLong()
+
+    val uploadKey = "claim${UUID.randomUUID()}$fileName"
```
palmenhq: didn't see this before but shouldn't we have some kind of separators? Also wouldn't it be better to have the claim id here also? something like `val uploadKey = "claim-$claimId/${UUID.randomUUID()}-$fileName"` or so

src/main/resources/schema.graphqls
```diff
@@ -553,11 +565,11 @@ enum ClaimTypes {
...
-  | AccidentalDamageClaim
-  | AssaultClaim
-  | WaterDamageClaim
-  | TravelAccidentClaim
-  | LuggageDelayClaim
+  | AccidentalDamageClaim
+  | AssaultClaim
+  | WaterDamageClaim
+  | TravelAccidentClaim
+  | LuggageDelayClaim
```
palmenhq: these f**** nbsps... very nice change!

PR #229, opened 2020-01-10
Feature/discount info hope

akashpatl:
Adding some more info under the account tab. This includes the total discount, total charge and what active discount code will be used for upcoming charge.

https://github.com/HedvigInsurance/account-service
https://github.com/HedvigInsurance/hedgehog

### DIFF
--- a/src/main/java/com/hedvig/backoffice/services/account/AccountServiceStub.java
+++ b/src/main/java/com/hedvig/backoffice/services/account/AccountServiceStub.java
@@ -4,6 +4,7 @@ import com.hedvig.backoffice.graphql.types.account.AccountEntryInput;
 import com.hedvig.backoffice.services.account.dto.*;
 import org.javamoney.moneta.Money;
 
+import javax.money.MonetaryAmount;
 import java.math.BigDecimal;
 import java.time.Instant;
 import java.time.LocalDate;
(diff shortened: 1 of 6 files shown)


src/main/java/com/hedvig/backoffice/services/account/AccountServiceStub.java
```diff
@@ -50,18 +58,23 @@ public AccountServiceStub() {
         Optional.empty()
       )
     );
+
+    chargeEstimation = new AccountChargeEstimationDTO(
+        Money.of(subscription, "SEK"),
+        Money.of(discount, "SEK"),
+          Money.of(currentMonthsBalance.add(subscription).subtract(discount), "SEK"),
+        ImmutableList.of("referral code")
```
palmenhq: any particular reason for making this an immutable list and not just `List#of`?

vonElfvin: Would be nice, in the future, if this could be the member-ids of the members that the member invited and the member id of the one that invited the member. This is fine for now though. A Kotlin list is already immutable as far as I am concerned so I think listOf("referral code") is better here.

akashpatl: My language level was not set correctly so `List.of` seemed unsupported, which of course it isn't.

akashpatl: 👍

PR #230, opened 2020-01-13
Use new create complete functionality from underwriter

johantjellden:
Depends on https://github.com/HedvigInsurance/underwriter/pull/60

### DIFF
--- a/src/main/java/com/hedvig/backoffice/services/underwriter/UnderwriterServiceImpl.kt
+++ b/src/main/java/com/hedvig/backoffice/services/underwriter/UnderwriterServiceImpl.kt
@@ -41,7 +41,7 @@ class UnderwriterServiceImpl(
         incompleteApartmentQuoteData = quoteDto.incompleteApartmentQuoteData?.let((IncompleteApartmentQuoteDataDto)::from),
         incompleteHouseQuoteData = quoteDto.incompleteHouseQuoteData?.let((IncompleteHouseQuoteDataDto)::from),
         quotingPartner = null,
-        complete = true,
+        shouldComplete = true,
         underwritingGuidelinesBypassedBy = null
     )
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/hedvig/backoffice/services/underwriter/UnderwriterServiceImpl.kt
```diff
@@ -42,22 +40,29 @@ class UnderwriterServiceImpl(
...
+  private fun underwriterCreateQuote(quoteRequestDto: QuoteRequestDto): QuoteResponseDto? {
+    return try {
+      underwriterClient.createQuote(
+        quoteRequestDto
+      )
+      } catch (e: FeignException) {
+      logger.error("Failed to complete quote", e)
+      null
+      }
+  }
```
palmenhq: formatting here looks a bit off

src/main/java/com/hedvig/backoffice/services/underwriter/dtos/QuoteRequestDto.kt
```diff
@@ -15,7 +15,9 @@ data class QuoteRequestDto(
   val incompleteHouseQuoteData: IncompleteHouseQuoteDataDto?,
   val incompleteApartmentQuoteData: IncompleteApartmentQuoteDataDto?,
   val memberId: String? = null,
-  val originatingProductId: UUID? = null
+  val originatingProductId: UUID? = null,
+  val complete: Boolean = true,
+  val underwritingGuidelinesBypassedBy: String?
```
palmenhq: any particular reason for removing the default `null` here?

johantjellden: Me and elvin discussed default parameters a while back and thought that many times it can be good to avoid them. In this specific case I though that the parameter is important enough to not have a default value. And since we use it at only a few places it doesn't really hurt to write it out.

Maybe I should move it further up in the parameter list thou.

PR #239, opened 2020-02-04
Actually restrict H.OPE. access

palmenhq:
Just wanted to make a point of how easy it becomes to restrict the access to hope. If we're fine with granting access to others via the gatekeeper DB for now, this is sufficient in a first step.


vonElfvin:
Cool! That does look nice :)

PR #248, opened 2020-03-25
Add ContractMarket to member in schema

vonElfvin:


src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -97,4 +98,6 @@ class MemberResolver(
     .map((Quote)::from)
 
   fun getContracts(member: Member): List<Contract> = productPricingService.getContractsByMemberId(member.memberId)
+
+  fun getContractMarket(member: Member): ContractMarketInfo? = productPricingService.getContractMarketInfoByMemberId(member.memberId)
```
AlexSJ1990: getContractMarketInfo ?

vonElfvin: The function is named after the property in the schema, could name it info in the schema however

PR #253, opened 2020-04-15
Switch account-service dtos to Kotlin, use MonetaryAmountV2 for graphql

vonElfvin:


src/main/resources/schema.graphqls
```diff
@@ -710,16 +710,16 @@ type NumberFailedCharges {
 
 type Account {
   id: ID!
-  currentBalance: MonetaryAmount!
-  totalBalance: MonetaryAmount!
+  currentBalance: MonetaryAmountV2!
```
FredrikAreschoug, line 713: Are you sure that this will not break the schema in a way that brakes the apps?

vonElfvin:
Before mergin this I need to have https://github.com/HedvigInsurance/hedgehog/pull/164 ready since they are dependent on each other for H.OPE. to work

PR #256, opened 2020-04-20
Implementation of itemizer-service

vonElfvin:


src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.kt
```diff
@@ -0,0 +1,119 @@
+package com.hedvig.backoffice.graphql.resolvers
```
FredrikAreschoug, line 1: I think kotlinizing this is fine but doing it in a pr with the life span of this one is problematic. For example, there are changes done in this class in #257

src/main/java/com/hedvig/backoffice/services/itemizer/ItemizerServiceImpl.kt
```diff
@@ -17,6 +17,7 @@ class ItemizerServiceImpl(
     ItemCategoryKind.COMPANY -> itemizerClient.getCompanies()
   }
 
+
```
FredrikAreschoug, line 20:
```suggestion
```
:monocle_face:

src/main/resources/schema.graphqls
```diff
@@ -94,7 +94,13 @@ type MutationType {
   upsertItemModel(request: UpsertItemModelInput): ID!
   upsertClaimItem(request: UpsertClaimItemInput): ID!
   deleteClaimItem(claimItemId: ID!): ID
+  addItemCategories(request: AddItemCategoriesInput): [Boolean!]!
 }
+
+input AddItemCategoriesInput {
+  itemCategoriesString: String!
```
FredrikAreschoug, line 111:
```suggestion
  itemCategories: String!
```
Don't know If I think the String part is needed

PR #257, opened 2020-04-28
send and fetch messages throught graphql

FredrikAreschoug:


src/main/java/com/hedvig/backoffice/services/chat/ChatServiceV2.kt
```diff
@@ -0,0 +1,10 @@
+package com.hedvig.backoffice.services.chat
+
+import com.hedvig.backoffice.services.chat.data.SendMessageResponse
+
+interface ChatServiceV2 {
```
vonElfvin, line 5: Any particular reason you did not use the old one? I would prefer us taking one step at a time so to speak. You could kotlinize it if that's what you are looking for.

vonElfvin: Since it works currently and we are making no major changes I'd prefer if we translated the new to the old for now and then look to completely refactoring once that works.

FredrikAreschoug: I V2 it because this way we could keep the old solution intact so if this new way would fail we have a fallback using V1. Of course, I could implement the new `fetchMessages` and `sendMessage` in V1 but it's just cluttered with stuff that we probably don't want. And I se no gain in kotlinize and stay clear of code like that. And this gives us the plus of just deleting V1 when it's tested.

FredrikAreschoug:
@vonElfvin still `Changes requested`?

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -117,25 +117,22 @@ public GraphQLQuery(
...
-
     return ticketService.getAllTickets(resolved);
   }
 
   public String me(DataFetchingEnvironment env) {
     try {
       return GraphQLConfiguration.getEmail(env, personnelService);
     } catch (Exception e) {
-      log.info("Exception occured when trying to access user email: " + e);
+      log.info("Exception occured when trying to access user getEmail: " + e);
```
FredrikAreschoug, line 135:
```suggestion
      log.info("Exception occured when trying to access user email: " + e);
```

src/main/java/com/hedvig/backoffice/graphql/types/ChatMessage.kt
```diff
@@ -1,5 +1,29 @@
...
+import java.time.Instant
+
 data class ChatMessage(
-  val messageAsJson: String
-)
+  val globalId: Long,
+  val author: String?,
+  val fromId: Long,
+  val timestamp: Instant?,
+  val messageBodyJsonString: String
```
FredrikAreschoug, line 14: :+1:

PR #261, opened 2020-05-18
Fetch many member with distinct list

FredrikAreschoug:
Member service will return 404 if the member id appears twice in the list :)


FredrikAreschoug:
When testing the switcher automation combos was not working maybe it should not have multiple entries of the same id in the first place, but it's something we found when adding multiple contracts on a member

PR #262, opened 2020-05-19
reverted distinct and converted to kotlin

FredrikAreschoug:
#261 was not a good solution


src/main/java/com/hedvig/backoffice/graphql/dataloaders/MemberBatchLoader.kt
```diff
@@ -0,0 +1,24 @@
...
+import java.util.stream.Collectors
+
+@Component
+class MemberBatchLoader(private val memberService: MemberService) : BatchLoader<String, Member> {
+
+  override fun load(keys: List<String>): CompletionStage<List<Member>> {
+    return CompletableFuture.supplyAsync {
+      memberService.getMembersByIds(keys)
+        .map { fromDTO(it) }
+        .sortedBy { keys.indexOf(it.memberId) }
```
meletios, line 21: Is this doing the same thing as distinct? or do we want something else?

FredrikAreschoug: No this dose, the same thing as `.sorted(Comparator.comparing(item -> keys.indexOf(item.getMemberId())))` this PR should only revert the changes I did in #261 but converted this class so felt like why let that go to waste :smile:

FredrikAreschoug: [Made a fix in member-service instead](https://github.com/HedvigInsurance/member-registry-service/pull/400)

PR #263, opened 2020-05-20
get campaign codes from product and pricing and add query for finding…

AlexSJ1990:
…CampaignCodes


src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -169,4 +170,11 @@ public String me(DataFetchingEnvironment env) {
       .map(SwitchableSwitcherEmail::from)
       .collect(Collectors.toList());
   }
+
+  public List<VoucherCampaign> findPartnerCampaigns() {
+    return productPricingService.searchPartnerCampaigns().stream().map( partnerCampaignSearchResponse ->
+      VoucherCampaign.Companion.from(partnerCampaignSearchResponse)
+    ).collect(Collectors.toList());
+  }
```
meletios, line 178: i have an objection with the naming here. I dont know why, but it feels more clear to me if its getCampaigns instead of search.

feel free to discard though

```suggestion
  public List<VoucherCampaign> getPartnerCampaigns() {
    return productPricingService.getPartnerCampaigns().stream().map( partnerCampaignSearchResponse ->
      VoucherCampaign.Companion.from(partnerCampaignSearchResponse)
    ).collect(Collectors.toList());
  }
```

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -169,4 +166,20 @@ public String me(DataFetchingEnvironment env) {
...
+  public List<VoucherCampaign> findPartnerCampaigns(CampaignFilter filter) {
+    final var code = filter == null ? null : filter.getCode();
+    final var partnerId = filter == null ? null : filter.getPartnerId();
+    final var activeFrom = filter == null ? null : filter.getActiveFrom();
+    final var activeTo = filter == null ? null : filter.getActiveTo();
+    return productPricingService.searchPartnerCampaigns(code, partnerId, activeFrom, activeTo)
+      .stream().map(partnerCampaignSearchResponse ->
+      VoucherCampaign.Companion.from(partnerCampaignSearchResponse)
+    ).collect(Collectors.toList());
+  }
```
vonElfvin, line 179: I'm not sure how large this list is and I could be totally off, but if the list is manageable this could have been performed in the frontend to reduce the amount of dtos/requests/etc. going back and forth throough back-office. This is good however and will scale better going forward when the amount increases.

PR #267, opened 2020-06-05
Generate payment link from api gateway

palmenhq:
Connected to https://github.com/HedvigInsurance/hedgehog/pull/181

Also gets tests running again after a small break 😬 and removes one test which seemed a bit weird


src/main/resources/application.yaml
```diff
@@ -115,6 +115,11 @@ gatekeeper:
   baseUrl: https://id.dev.hedvigit.com
   stub: false
 
+api-gateway:
+  baseUrl: http://localhost:8080
+  stub: false
+  token: blah
```
FredrikAreschoug, line 121: :monocle_face:

palmenhq: we need to but something 🤷‍♀️ i could also put `your token here` but how fun is that ;)

PR #268, opened 2020-06-12
readd deleted queries

AlexSJ1990:


src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -148,4 +145,20 @@ public String me(DataFetchingEnvironment env) {
   public List<ClaimItem> claimItems(UUID claimId) {
     return itemizerService.getClaimItems(claimId);
   }
+
+  public List<VoucherCampaign> findPartnerCampaigns(CampaignFilter filter) {
+    final var code = filter == null ? null : filter.getCode();
+    final var partnerId = filter == null ? null : filter.getPartnerId();
+    final var activeFrom = filter == null ? null : filter.getActiveFrom();
+    final var activeTo = filter == null ? null : filter.getActiveTo();
```
meletios, line 153: Dont get me wrong, i do like elvis operator, but is it better if we say
if (filter == null) return emptylist and then continue with the search?

meletios: You can do it with an elvis operator as well 😂

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -148,4 +145,20 @@ public String me(DataFetchingEnvironment env) {
...
+    return productPricingService.searchPartnerCampaigns(code, partnerId, activeFrom, activeTo)
+      .stream().map(partnerCampaignSearchResponse ->
+        VoucherCampaign.Companion.from(partnerCampaignSearchResponse)
+      ).collect(Collectors.toList());
```
meletios, line 157:
```suggestion
    return  filter == null ? emptyList() : productPricingService.searchPartnerCampaigns(code, partnerId, activeFrom, activeTo)
      .stream().map(partnerCampaignSearchResponse ->
        VoucherCampaign.Companion.from(partnerCampaignSearchResponse)
      ).collect(Collectors.toList());
```

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
palmenhq: this is java though so i guess that'd require converting this whole blob to kotlin?

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
palmenhq: does it return none or all, though?

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
meletios: Aha! But we can use `new ArrayList` or something equivalent right?

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
meletios: Good question! Cause my comment is valid only in that case!

palmenhq: I think we should leave the search logic for product pricing, so imo this implementation is good

vonElfvin, line 157: Imo this would be neat, remove all the variables at the top
```suggestion
    final val campaigns = filter == null ?  productPricingService.searchPartnerCampaigns(null, null, null, null) : productPricingService.searchPartnerCampaigns(filter.getCode(), filter.getPartnerId(), filter.getActiveFrom(), filter.getActiveTo())
    return campaigns.stream().map(partnerCampaignSearchResponse ->
        VoucherCampaign.Companion.from(partnerCampaignSearchResponse)
      ).collect(Collectors.toList());
```

PR #272, opened 2020-06-26
Only update chat when asked for by the app

vonElfvin:


vonElfvin:
This does not seem prio atm

PR #273, opened 2020-06-29
Feature/add contract id to claim

AlexSJ1990:
related to pr - https://github.com/HedvigInsurance/claims-service/pull/67
enable contractId to be added to a claim via a dropdown in hope.

### DIFF
--- a/src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
+++ b/src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
@@ -8,7 +8,7 @@ import com.hedvig.backoffice.graphql.dataloaders.MemberLoader;
 import com.hedvig.backoffice.graphql.types.Claim;
 import com.hedvig.backoffice.graphql.types.*;
 import com.hedvig.backoffice.graphql.types.account.AccountEntryInput;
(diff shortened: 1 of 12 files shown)


src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
```diff
@@ -728,6 +729,12 @@ public Boolean assignCampaignToPartnerPercentageDiscount(AssignVoucherPercentage
     return true;
   }
 
+  public Boolean addContractIdToClaim(AddContractIdToClaim request, DataFetchingEnvironment env) {
+    claimsService.addContractIdToClaim(request);
```
AlexSJ1990, line 733: do we want to store who added the contractId to the claim?

src/main/java/com/hedvig/backoffice/graphql/types/claims/SetContractForClaim.kt
```diff
@@ -0,0 +1,7 @@
+package com.hedvig.backoffice.graphql.types.claims
+
+data class SetContractForClaim(
+  val claimId: String,
+  val memberId: String,
```
vonElfvin, line 5: The memberId is most likely already on the claim itself (the ClaimAggregate) so no need to send it. But it is fine

PR #274, opened 2020-06-30
Implementation of automated depreciation and valuation

rasmusguterstam:
Implementation of https://github.com/HedvigInsurance/hedgehog/pull/196


src/main/resources/schema.graphqls
```diff
@@ -135,9 +149,31 @@ type ClaimItem {
   itemCompany: ItemCompany
   dateOfPurchase: LocalDate
   purchasePrice: MonetaryAmountV2
+  valuation: MonetaryAmountV2
   note: String
 }
 
+type Evaluation {
```
palmenhq, line 156: just checking, isn't `evaluation` a bit of a weird name? (translates to "utvärdering" it swedish no?) wouldn't `valuation` be more appropriate?

palmenhq: Also, since graphql doesn't offer any namespacing `valuation`/`evaluation` might be a bit generic for the use case, shall we prefix it with something like `ClaimItemValuation` or so?

src/main/resources/schema.graphqls
```diff
@@ -135,9 +149,31 @@ type ClaimItem {
...
+  depreciatedValue: Float
```
palmenhq, line 157: `MonetaryAmount` no? Floats are really not appropriate for money

palmenhq: or if its some kind of ratio rather i think we should use integers instead, for example a scale of 0-100 or 0-1000. Floats are way too insecure imo because of rounding errors.

src/main/resources/schema.graphqls
```diff
@@ -135,9 +149,31 @@ type ClaimItem {
...
+  evaluationRule: EvaluationRule
+}
+
+type EvaluationRule {
+  evaluationName: String!
+  itemFamily: String!
+  itemTypeId: ID
+  ageLimit: Float!
```
palmenhq, line 165: Same here re rounding errors, can we use ints instead?

src/main/resources/schema.graphqls
```diff
@@ -135,9 +149,31 @@ type ClaimItem {
...
+  typeOfContract: String!
+  evaluationType: String!
+  depreciation: Float
```
palmenhq, line 168: aand you know the drill

src/main/resources/schema.graphqls
```diff
@@ -148,6 +184,9 @@ input UpsertClaimItemInput {
   dateOfPurchase: LocalDate
   purchasePriceAmount: Float
   purchasePriceCurrency: String
+  valuationAmount: Float
+  customValuationAmount: Float
+  valuationCurrency: String
```
palmenhq, line 189: i think we should use `monetaryAmount` here for sure

src/main/resources/schema.graphqls
rasmusguterstam: @palmenhq In this context, ageLimit is just informative of which age bracket the depreciation comes from ('...since this item is at least 1.5 years old'), and it is stored as a BigDecimal in the actual service. It is even reduced to 2 digits in the service before returning it, so isn't it justifiable to use a Float just here? 😬 Feels weird to use ints when it's not a ratio?

palmenhq: TLDR; Right! Carry on 👜 

Reasoning:
I see your issue. Normally I'd suggest using fractionized integers instead - especially considering we know the number of decimals we need. For example `1.5` years would be represented as `150` years until the very last step in the view where it's divided by `100`. This is a fairly common approach when not having access to arbitrary-precision data types such as `BigDecimal` (like we don't in for example JS or pre Java 7). But considering we're not doing (and shouldn't be doing) any other math on the frontend, and this number might not need to be super-exact, and additionally is backed by `BigDecimal`, i guess this is ok :) But in general we should be very careful about floating point numbers.

PR #276, opened 2020-07-02
Feature/question group country sort

gustafvh:
Add Resolver for QuestionGroup, so that it contains member, which in turn contains Contract Market info.


src/main/java/com/hedvig/backoffice/graphql/resolvers/QuestionGroupResolver.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.stereotype.Component
+import java.time.Instant
+
+@Component
+class QuestionGroupResolver
+  (
+  private val memberService: MemberService,
+  private val personnelService: PersonnelService
+) :
+  GraphQLResolver<QuestionGroupType?> {
```
vonElfvin, line 21:
```suggestion
) : GraphQLResolver<QuestionGroupType?> {
```

src/main/java/com/hedvig/backoffice/graphql/resolvers/QuestionGroupResolver.kt
```diff
@@ -0,0 +1,50 @@
...
+        "Unknown",
+        Gender.OTHER,
+        "Unknown",
+        "Unknown",
+        "Unknown",
+        Instant.now(),
+        "Unknown",
+        "Unknown"
+      )
+    }
```
vonElfvin, line 47: The try catch is unnecessary here since it is will most likely never happen:
```suggestion
      fromDTO(
        memberService.findByMemberId(
          QuestionGroup.memberId,
          GraphQLConfiguration.getIdToken(env, personnelService)
        )
      )
```

vonElfvin: Also, for code clartiy, we should not import static companion functions like this:
`fromDTO` should be `Member.fromDTO`

gustafvh: It does cause a exception-crash without it, on the data I currently have. Are we sure its not just best to keep it in for those few times it does happen?

vonElfvin: That's because you are stubbing member-service. With correct data accross the services that should not happen. You could also fix the member-service stub to not return nothing if a member is not found.

src/main/java/com/hedvig/backoffice/graphql/resolvers/QuestionGroupResolver.kt
```diff
@@ -0,0 +1,34 @@
...
+    return try {
+      Member.fromDTO(
+        memberService.findByMemberId(
+          QuestionGroup.memberId,
+          GraphQLConfiguration.getIdToken(env, personnelService)
+        )
+      )
+    } catch (e: Exception) {
+      null
+    }
```
vonElfvin, line 31: Here we should log and capture the exception to Sentry
```suggestion
    } catch (exception: Exception) {
      logger.error("Unable get to resolve member for Question Group (memberId=${questionGroup.memberId}), exception)
      Sentry.capture(exception)
      null
    }
```

src/main/java/com/hedvig/backoffice/graphql/resolvers/QuestionGroupResolver.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.stereotype.Component
+
+@Component
+class QuestionGroupResolver
+  (
+  private val memberService: MemberService,
+  private val personnelService: PersonnelService
+) : GraphQLResolver<QuestionGroupType?> {
+  fun getMember(
+    QuestionGroup: QuestionGroupType,
```
vonElfvin, line 19: Camelcase on variables
```suggestion
    questionGroup: QuestionGroupType,
```

src/main/java/com/hedvig/backoffice/graphql/resolvers/QuestionGroupResolver.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.hedvig.backoffice.services.personnel.PersonnelService
+import graphql.schema.DataFetchingEnvironment
+import org.springframework.stereotype.Component
+
+@Component
+class QuestionGroupResolver
+  (
+  private val memberService: MemberService,
+  private val personnelService: PersonnelService
+) : GraphQLResolver<QuestionGroupType?> {
```
vonElfvin, line 17:
```suggestion
) : GraphQLResolver<QuestionGroupType> {
```

src/main/java/com/hedvig/backoffice/graphql/resolvers/QuestionGroupResolver.kt
```diff
@@ -0,0 +1,44 @@
...
+      )
+    } catch (exception: Exception) {
+      logger.error("Unable get to resolve member for Question Group (memberId=${questionGroup.memberId}), exception");
+      Sentry.capture(exception)
+      null
+    }
+  }
+
+  companion object {
+    private val logger = LoggerFactory.getLogger(ChatServiceV2Impl::class.java)
```
vonElfvin, line 41:
```suggestion
    private val logger = LoggerFactory.getLogger(QuestionGroupResolver::class.java)
```

PR #277, opened 2020-07-03
Pass along the actual message to notification-service when responding to a chat message

hedvigoscar:
This is done to enable Android to have Messaging-style notifications later on.

Note that I'm having trouble testing this feature entirely as getting the back-office application stack running locally has proven quite challenging

### DIFF
--- a/src/main/java/com/hedvig/backoffice/services/chat/ChatServiceV2.kt
+++ b/src/main/java/com/hedvig/backoffice/services/chat/ChatServiceV2.kt
@@ -6,6 +6,6 @@ import com.hedvig.backoffice.services.messages.dto.BotMessageDTO
 interface ChatServiceV2 {
   fun fetchMessages(memberId: String, personnelEmail: String, personnelToken: String): List<BotMessageDTO>
   fun sendMessage(memberId: String, message: String, forceSendMessage: Boolean, personnelEmail: String, personnelToken: String): SendMessageResponse
-  fun sendNotification(memberId: String, personnelToken: String)
+  fun sendNotification(memberId: String, personnelToken: String, message: String)
 }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/hedvig/backoffice/services/chat/ChatServiceV2.kt
```diff
@@ -6,6 +6,6 @@ import com.hedvig.backoffice.services.messages.dto.BotMessageDTO
 interface ChatServiceV2 {
   fun fetchMessages(memberId: String, personnelEmail: String, personnelToken: String): List<BotMessageDTO>
   fun sendMessage(memberId: String, message: String, forceSendMessage: Boolean, personnelEmail: String, personnelToken: String): SendMessageResponse
-  fun sendNotification(memberId: String, personnelToken: String)
+  fun sendNotification(memberId: String, personnelToken: String, message: String)
```
vonElfvin, line 9: Is `message` the message of the notification or the message that has been sent from IEX in the chat?

vonElfvin: Also, what happens if the message is null?

vonElfvin: If this is a very "chat specific" notification maybe we should consider renaming the functions to "sendChatMessageSentNotification" or similar so it is clear what it does

hedvigoscar: Message is actually both - after this change is fully implemented, the notification will contain the IEX message.

It can't be null, Kotlin enforces this

Naming was off but is now improved 🙂

PR #279, opened 2020-07-10
Implementation of Referral queries & mutations

rasmusguterstam:
Back-office part of [Notion ticket](https://www.notion.so/Tech-2d70066c6dc44d2a929bfc34c13e381e?p=ef779a89eff5440fa743d13af40a56c0)

**Implements the following end-points of product-pricing:**

- `manualRedeemEnableReferralsCampaign`
- `manualRedeemCampaign`
- `manualUnRedeemCampaign`
- `eligibleForReferral`
- `referralInformation`

### DIFF
--- a/src/main/java/com/hedvig/backoffice/services/product_pricing/dto/ManualRedeemCampaignRequest.kt
+++ b/src/main/java/com/hedvig/backoffice/services/product_pricing/dto/ManualRedeemCampaignRequest.kt
@@ -4,5 +4,5 @@ import java.time.LocalDate
 
 data class ManualRedeemCampaignRequest (
   val campaignCode: String,
-  val activationDate: LocalDate
+  val activationDate: LocalDate?
 )
--- a/src/main/java/com/hedvig/backoffice/services/product_pricing/dto/ManualRedeemCampaignRequest.kt
+++ b/src/main/java/com/hedvig/backoffice/services/product_pricing/dto/ManualRedeemCampaignRequest.kt
@@ -1,0 +1,8 @@
+package com.hedvig.backoffice.services.product_pricing.dto
+
+import java.time.LocalDate
+
+data class ManualRedeemCampaignRequest (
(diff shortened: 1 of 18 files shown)


src/main/resources/schema.graphqls
```diff
@@ -120,6 +120,9 @@ type MutationType {
   setContractForClaim(
     request: SetContractForClaim!
   ): Boolean!
+  manualRedeemCampaign(memberId: ID!, request: ManualRedeemCampaignInput!): Boolean!
+  manualUnRedeemCampaign(memberId: ID!, request: ManualUnRedeemCampaignInput!): Boolean!
+  manualRedeemEnableReferralsCampaign(market: Market!, request: ManualRedeemEnableReferralsCampaignInput): Boolean!
```
FredrikAreschoug, line 125: Will we have the possibility to batch enable members in hope? otherwise, we could go  with just:
```suggestion
  manualRedeemEnableReferralsCampaign(memberId: ID!, market: Market!): Boolean!
```
Then we could use the batch enpoint end send a list of one memberId

rasmusguterstam: Good point! 😀

src/main/java/com/hedvig/backoffice/services/product_pricing/dto/ManualRedeemCampaignRequest.kt
```diff
@@ -0,0 +1,8 @@
+package com.hedvig.backoffice.services.product_pricing.dto
+
+import java.time.LocalDate
+
+data class ManualRedeemCampaignRequest (
+  val campaignCode: String,
+  val activationDate: LocalDate
```
FredrikAreschoug, line 7: `activationDate` is nullable and send it as null if the hope operator is not specifically set in it to something
```suggestion
  val activationDate: LocalDate?
```

PR #280, opened 2020-07-13
Add ability to get a member's PickedLocale in graphQL

vonElfvin:


src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -3,23 +3,16 @@ package com.hedvig.backoffice.graphql.resolvers
...
 import com.hedvig.backoffice.graphql.dataloaders.AccountLoader
-import com.hedvig.backoffice.graphql.types.DirectDebitStatus
-import com.hedvig.backoffice.graphql.types.FileUpload
-import com.hedvig.backoffice.graphql.types.Member
-import com.hedvig.backoffice.graphql.types.MonthlySubscription
-import com.hedvig.backoffice.graphql.types.Person
-import com.hedvig.backoffice.graphql.types.ProductType
-import com.hedvig.backoffice.graphql.types.Quote
-import com.hedvig.backoffice.graphql.types.Transaction
+import com.hedvig.backoffice.graphql.types.*
```
meletios, line 6: 😧

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -31,7 +24,7 @@ import graphql.schema.DataFetchingEnvironment
 import io.sentry.Sentry
 import org.springframework.stereotype.Component
 import java.time.YearMonth
-import java.util.ArrayList
+import java.util.*
```
meletios, line 27: 😧

src/main/java/com/hedvig/backoffice/services/members/MemberService.kt
```diff
@@ -1,41 +1,36 @@
 package com.hedvig.backoffice.services.members
 
-import com.hedvig.backoffice.services.members.dto.InsuranceCancellationDTO
-import com.hedvig.backoffice.services.members.dto.MemberDTO
-import com.hedvig.backoffice.services.members.dto.MembersSearchResultDTO
-import com.hedvig.backoffice.services.members.dto.MembersSortColumn
-import com.hedvig.backoffice.services.members.dto.PersonDTO
-import com.hedvig.backoffice.services.members.dto.PickedLocaleDTO
+import com.hedvig.backoffice.services.members.dto.*
```
meletios, line 3: 😧

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -179,6 +172,7 @@ class MemberResolver(
     return contracts
   }
 
-  fun getContractMarketInfo(member: Member): ContractMarketInfo? =
-    productPricingService.getContractMarketInfoByMemberId(member.memberId)
+  fun getContractMarketInfo(member: Member): ContractMarketInfo? = productPricingService.getContractMarketInfoByMemberId(member.memberId)
+
+  fun getPickedLocale(member: Member): PickedLocale = memberService.findPickedLocaleByMemberId(member.memberId).pickedLocale
```
meletios, line 177: Do we always have pickedLocale? Should this be nullable?

src/main/resources/schema.graphqls
```diff
@@ -653,6 +653,7 @@ type Member {
   quotes: [Quote!]!
   contracts: [Contract!]!
   contractMarketInfo: ContractMarketInfo
+  pickedLocale: PickedLocale
```
FredrikAreschoug, line 656: A suggestion is to have this as a string in back-office then we would have to updateit with new markets. We could just map the strings in the front end

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
vonElfvin: Since the DTO assigned it as non-nullable I assume it is never null.

src/main/resources/schema.graphqls
vonElfvin: I started off with a string, but since hedgehog uses typescript which is typed I thought having enums is pretty neat. It will be annoying if we add another PickedLocale option but it does give clarify. What do you think I should go with?

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
vonElfvin: option+cmd+L 🤷 Is there a way to configure it to do not to this?

src/main/resources/schema.graphqls
FredrikAreschoug: this is defiantly clearer what you expect, it's really hard when back-office sometimes is just a middleman and sometimes actually dose some logic. But I would go with String one less thing to change whenever we go to a new market :)

PR #284, opened 2020-07-20
Revert "Update member type for graphql"

FredrikAreschoug:
Reverts HedvigInsurance/back-office#281


FredrikAreschoug:
this is causing issues in the calims lets do a revert and add them again when hope is up to speed

PR #289, opened 2020-07-28
Implement QualityAssuranceService

rasmusguterstam:
Implements the service QualityAssurance which provides staging specific features (as of now, only unsignMember)


src/main/resources/schema.graphqls
```diff
@@ -123,6 +123,7 @@ type MutationType {
   manualRedeemCampaign(memberId: ID!, request: ManualRedeemCampaignInput!): Boolean!
   manualUnRedeemCampaign(memberId: ID!, request: ManualUnRedeemCampaignInput!): Boolean!
   manualRedeemEnableReferralsCampaign(memberId: ID!, market: Market!): Boolean!
+  unsignMember(market: String!, ssn: String!): Boolean
```
FredrikAreschoug, line 126: We should be able to guarantee that we can return a non null boolean :) 
```suggestion
  unsignMember(market: String!, ssn: String!): Boolean!
```

PR #300, opened 2020-09-15
Feature/generic agreements

AlexSJ1990:


src/main/resources/schema.graphqls
```diff
@@ -265,7 +265,7 @@ type Contract {
   terminationDate: LocalDate
   currentAgreementId: ID!
   hasPendingAgreement: Boolean!
-  agreements: [Agreement!]!
+  genericAgreements: [GenericAgreement!]!
```
vonElfvin, line 268: I'd rather keep the name "agreement" and have it called GenericAgreement in Kotlin, from graphQL perspective it is sufficient with Agreement. Or what do you think?

vonElfvin (commented):
Clean! Added a thought

src/main/resources/schema.graphqls
vonElfvin: Nvm, I changed my mind. Keep it as GenericAgreement

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -121,7 +121,7 @@ class MemberResolver(
     return reqularAndSignableQuotes(member.memberId, quotes)
   }
 
-
+// todo: fix
```
vonElfvin, line 124: Fix it

vonElfvin, line 124: Maybe use a bit more words to explain why this should be fixed

vonElfvin (approved):
Clean

vonElfvin (approved):

PR #306, opened 2020-09-23
Json schema resolver

johantjellden:


pom.xml
```diff
@@ -29,6 +29,7 @@
     <mockito.version>2.22.0</mockito.version>
     <kotlin.version>1.3.31</kotlin.version>
     <kotlin.compiler.incremental>true</kotlin.compiler.incremental>
+    <byte-buddy.version>1.9.3</byte-buddy.version>
```
vonElfvin, line 32: Don't think this is used?

src/main/java/com/hedvig/backoffice/graphql/resolvers/QuoteSchemaResolver.kt
```diff
@@ -0,0 +1,13 @@
+package com.hedvig.backoffice.graphql.resolvers
+
+import com.coxautodev.graphql.tools.GraphQLResolver
+import com.hedvig.backoffice.graphql.types.Quote
+import com.hedvig.backoffice.services.underwriter.UnderwriterService
+
+class QuoteSchemaResolver(
```
vonElfvin, line 7: Resolvers are on "the parent" not what they resolve afaik
```suggestion
class QuoteResolver(
```

src/main/java/com/hedvig/backoffice/services/underwriter/UnderwriterServiceImpl.kt
```diff
@@ -191,4 +192,15 @@ class UnderwriterServiceImpl(
     }
     throw IllegalStateException("Cannot sign quote [QuoteId: $completeQuoteId]")
   }
+
+  override fun getSchemaFromQuote(id: UUID): JsonNode?  =
+    try {
+      underwriterClient.getSchemaFromQuote(id)!!.body
+    }catch (ex: RestClientResponseException) {
+      if(ex.rawStatusCode == 404) {
```
vonElfvin, line 200:
```suggestion
      if (ex.rawStatusCode == 404) {
```

vonElfvin (commented):
Looks neat, nice with some tests as well :)

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
```diff
@@ -203,6 +209,10 @@ public CanValuateClaimItem canValuateClaimItem(TypeOfContract typeOfContract, St
     return itemizerService.canValuateClaimItem(typeOfContract, itemFamilyId, itemTypeId);
   }
 
+  public JsonNode getQuoteSchemaForContractType(ContractType contractType) {
```
FredrikAreschoug, line 212: Why don't we go with a string in this case? I think it's going to be nicer to just threat Back office as a proxy

FredrikAreschoug (commented):
Overall it looks good just added a question

src/main/java/com/hedvig/backoffice/graphql/resolvers/GraphQLQuery.java
vonElfvin: It was a string initially. It was just pretty neat to have the enum available in hedgehog 🤔

johantjellden: Aha

johantjellden: Yes i thought about that as well to not have the contract constants.

FredrikAreschoug: But hedgehog can treat is as an enum it's just but since it's an internal graphql schema I think it's just going to bite us in the as having it as an enum. If it were an open schema my opinion would be the opposite

vonElfvin: How do you mean with treating it as an enum? If it is a string then I don't have an enum?

FredrikAreschoug: Implement it as an enum in js and send it as a String through graphql

FredrikAreschoug (approved):
🖖

PR #324, opened 2020-12-03
[HVG-623] Enable Denmark in back-office

vonElfvin:
Why:
- Currently back-office does not support the functionality expected in Denmark
- Make back-office more market agnostic
- Some boyscouting on the way

What:
- Make Market, QuoteProductType and PickedLocale into string for back-office
- Clean up enable referrals
- Make quotes signable in Denmark


src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
```diff
@@ -136,7 +135,8 @@ class MemberResolver(
         return reqularAndSignableQuotes(member.memberId, quotes)
     }
 
-    // todo: fix do not use contractTypeName for logic here
+    // TODO: fix do not use contractTypeName for logic here
+    // FIXME: this should probably not live here in the first place
```
johantjellden, line 139: What is it we are doing here? 
Seems like we are checking if the purpose of the quote is to move to a new contract type?
Should this contract based logic be moved into underwriter?

AlexSJ1990 (approved):
🧹

src/main/java/com/hedvig/backoffice/graphql/resolvers/MemberResolver.kt
vonElfvin: We're checking if the quote should be signable or not. Should probably live in underwriter in the future yes I agree. For now I won't fix that though.

PR #325, opened 2020-12-03
Get rid of tickets

vonElfvin:
Tickets are no longer used in hope at all so may as well not have it in back-office


rasmusguterstam (approved):
🌟

PR #328, opened 2021-01-15
Correctly return null if no NumberFailedCharges found in account-service

vonElfvin:
## What?
- Return null if no member/number failed charges found in account-service

## Why?
- So we can safely ask for numberFailedCharges for non-signed members/those without an account


FredrikAreschoug (approved):
Nice

AlexSJ1990 (approved):
👍

PR #329, opened 2021-01-20
Catch 404 exception instead of checking uncheckable status

vonElfvin:
## What?
- Catch 404 exception instead of checking the status

## Why?
- Because the exception is thrown instead of getting the ResponseEntity.

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally


meletios (approved):
sweet

FredrikAreschoug (approved):

PR #333, opened 2021-02-04
experimental attempt at an auto-discovery mode of upstream services, handy for local dev

bystam:
# Jira Issue: - 

## What?
Replaces the `stub: true|false` configuration of upstream services with a `mode: live|stub|auto-discover`, where the `auto-discover` mode is the new one. It makes back-office choose between using a stubbed one or not depending on whether it can detect the service to be running, using its health endpoint.

This also moves the "mode" configuration to the regular `application.yml` under the development/test profiles, where is `auto-discover` by default.

Note: If the `mode` value is missing, it defaults to `live`.

## Why?
Manually changing `application-development.yml` files in the middle of development can be tedious, and it's easy to forget that setting from earlier development. This would make it work simply based on which other services you are running, which might be closer to what you want.

## Example startup output
When already running underwriter and claims and with `mode: auto-discover`, this is what it prints:
image

```
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [accountService]
... c.h.b.config.ExternalServicesConfig      : Did not discover service [accountService] - will use stub
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [claims]
... c.h.b.config.ExternalServicesConfig      : Discovered service [claims] - will use live one
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [memberservice]
... c.h.b.config.ExternalServicesConfig      : Did not discover service [memberservice] - will use stub
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [paymentService]
... c.h.b.config.ExternalServicesConfig      : Did not discover service [paymentService] - will use stub
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [botservice]
... c.h.b.config.ExternalServicesConfig      : Did not discover service [botservice] - will use stub
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [expo]
... c.h.b.config.ExternalServicesConfig      : Did not discover service [expo] - will use stub
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [underwriter]
... c.h.b.config.ExternalServicesConfig      : Discovered service [underwriter] - will use live one
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [productPricing]
... c.h.b.config.ExternalServicesConfig      : Did not discover service [productPricing] - will use stub
... c.h.b.config.ExternalServicesConfig      : Auto discovering upstream service [meerkat]
... c.h.b.config.ExternalServicesConfig      : Did not discover service [meerkat] - will use stub
```

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally


vonElfvin (commented):
Super cool idea!

Some things to consider:
1. May as well make this into Kotlin if we're changing the whole file
2. This assumes all services uses `/actuator/health`. Which I'm not 100% sure all of them do, Meerkat is one that might be using something different for example.
3. One cool condition would be that **if** it did sometimes run live, it should not switch back to stub. That way you can restart your services without risking getting weird behaviour.

bystam:
> Super cool idea!
> 
> Some things to consider:
> 
> 1. May as well make this into Kotlin if we're changing the whole file
> 2. This assumes all services uses `/actuator/health`. Which I'm not 100% sure all of them do, Meerkat is one that might be using something different for example.
> 3. One cool condition would be that **if** it did sometimes run live, it should not switch back to stub. That way you can restart your services without risking getting weird behaviour.

1. Absolutely
2. Yeah I'm aware - was going to refine such things if people felt like this idea was worthwhile ^^
3. That's already the case, no? This is all evaluated once during back-office launch, and never changed beyond that. So if for instance `claims` is restarted subsequently, back-office will just assume it's live still. It won't however support going from stub->live *after the fact*. That would require a lot more smart engineering hehe.

src/main/resources/application.yaml
```diff
@@ -44,66 +44,10 @@ hystrix:
...
-  tracker: 5000
   chat: 1000
-  claims: 60000
-
-tracker:
-  baseUrl: http://localhost:4082
-  urls:
-    assets: /assets
-    update: /asset
-  stub: false
```
bystam, line 56: I looked around the project and found no references to `tracker` at all, and then I saw that that repo hasn't been touched for 3 years, so I am thinking that maybe it's not used anymore?

src/main/resources/application.yaml
```diff
@@ -44,66 +44,10 @@ hystrix:
           enabled: false
 
 intervals:
-  tracker: 5000
   chat: 1000
-  claims: 60000
```
bystam, line 49: The `tracker` and `claims` values here appeared to be unused.

Michael-Duivestein (commented):
Nice!

src/main/resources/application.yaml
johantjellden: Yes poor asset-tracker never made it to launch. Remove it :)

src/main/java/com/hedvig/backoffice/config/ExternalServicesConfig.kt
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Bean
+    fun expoNotificationService(): ExpoNotificationService {
+        return create(
+            prefix = "expo",
+            stubClass = ExpoNotificationServiceStub::class,
+            liveClass = ExpoNotificationServiceImpl::class
+        )
+    }
```
johantjellden, line 98: This can most likely be removed as well. Expo was the first react framework we used to develop our apps. It had an framework specific way to send notifications to the apps.

If the implementation of this service still use expo we should remove it. 
@sampettersson @hedvigoscar do we still support exponotificaitons?

johantjellden (commented):
Left some comments regarding expo :)

src/main/java/com/hedvig/backoffice/config/ExternalServicesConfig.kt
hedvigoscar: I don't think we need to support it any more, no 😄

bystam: Updated!

FredrikAreschoug (approved):
I feel like people have had their say on this one let's get it out!

PR #335, opened 2021-02-09
improve the dockerfile to be in charge of building and testing too, not just image assembly

bystam:
# Jira Issue: [HVG-902]

## What?
This makes the back-office `Dockerfile` able to not only assemble the image, but also **build and test it**. This enables us to have a non-java non-maven specific pipeline in Codefresh that can be reused for any service. 

## Why?
This makes wiring a service to Codefresh very simple, and makes the "how to build and test" a service part of the git repo.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally

[HVG-902]: https://hedvig.atlassian.net/browse/HVG-902

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,18 +6,13 @@ COPY pom.xml /usr/src/app/
 RUN mvn -f /usr/src/app/pom.xml dependency:go-offline
 
 
-##### Compile stage #####
+##### Build stage #####
 FROM dependencies AS build
 
 COPY src/main /usr/src/app/src/main
 RUN mvn -f /usr/src/app/pom.xml clean package
 
-
-##### Test stage #####
-FROM dependencies AS test
-
 COPY src/test /usr/src/app/src/test
-COPY --from=build /usr/src/app/target /usr/src/app/target
 RUN mvn -f /usr/src/app/pom.xml test
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,12 @@ FROM dependencies AS build
 COPY src/main /usr/src/app/src/main
 RUN mvn -f /usr/src/app/pom.xml clean package
 
+
+##### Test stage #####
+FROM dependencies AS test
+
 COPY src/test /usr/src/app/src/test
+COPY --from=build /usr/src/app/target /usr/src/app/target
 RUN mvn -f /usr/src/app/pom.xml test
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,35 +7,28 @@ RUN mvn -f /usr/src/app/pom.xml dependency:go-offline
 
 
 ##### Compile stage #####
-FROM dependencies AS compile
+FROM dependencies AS build
 
 COPY src/main /usr/src/app/src/main
-RUN mvn -f /usr/src/app/pom.xml clean compile
+RUN mvn -f /usr/src/app/pom.xml clean package
 
 
 ##### Test stage #####
 FROM dependencies AS test
 
 COPY src/test /usr/src/app/src/test
-COPY --from=compile /usr/src/app/target /usr/src/app/target
+COPY --from=build /usr/src/app/target /usr/src/app/target
 RUN mvn -f /usr/src/app/pom.xml test
 
 
-##### Package stage #####
-FROM dependencies AS package
-
-COPY --from=compile /usr/src/app/target /usr/src/app/target
-RUN mvn -f /usr/src/app/pom.xml package
-
-
-##### Package artifact as runnable image #####
-FROM amazoncorretto:11 AS make_image
+##### Assemble artifact #####
+FROM amazoncorretto:11 AS assemble
 
 # Not sure what this one does
 RUN curl -o dd-java-agent.jar -L 'https://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=com.datadoghq&a=dd-java-agent&v=LATEST'
 
 # Copy the jar from build stage to this one
-COPY --from=package /usr/src/app/target/back-office-0.0.1-SNAPSHOT.jar /usr/app/
+COPY --from=build /usr/src/app/target/back-office-0.0.1-SNAPSHOT.jar /usr/app/
 
 # Define entry point
 ENTRYPOINT java -javaagent:/dd-java-agent.jar -jar /usr/app/back-office-0.0.1-SNAPSHOT.jar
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,7 +1,22 @@
+
+##### Build stage #####
+FROM maven:3.6.3-amazoncorretto-11 AS build
+
+# Copy source folder and pom.xml to this stage
+COPY src /usr/src/app/src
+COPY pom.xml /usr/src/app/
```
bystam, line 7: `/usr/src/app/` seems to be some kind of Dockerfile convention according to ze internets

Dockerfile
```diff
@@ -1,7 +1,22 @@
...
+
+# Build and produce an artifact
+RUN mvn -f /usr/src/app/pom.xml clean package -Dmaven.javadoc.skip=true -B -V -e
```
bystam, line 10: No `-DskipTests=true` here to simply make it run the tests immediately. We should look into whether `-Dmaven.repo.local=/codefresh/volume/m2_repository` should be needed here though

Dockerfile
```diff
@@ -1,7 +1,22 @@
...
+# Not sure what this one does
 RUN curl -o dd-java-agent.jar -L 'https://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=com.datadoghq&a=dd-java-agent&v=LATEST'
 
-ADD target/back-office-0.0.1-SNAPSHOT.jar /
+# Copy the jar from build stage to this one
+COPY --from=build /usr/src/app/target/back-office-0.0.1-SNAPSHOT.jar /usr/app/
 
-ENTRYPOINT java -javaagent:/dd-java-agent.jar -jar back-office-0.0.1-SNAPSHOT.jar
+# Define entry point
+ENTRYPOINT java -javaagent:/dd-java-agent.jar -jar /usr/app/back-office-0.0.1-SNAPSHOT.jar
```
bystam, line 22: Maybe some of this is a bit over-documented. I just found it helpful for me while writing it, and could be worthwhile when reproducing elsewhere :)

FredrikAreschoug (approved):
Wow cool stuff! :tada:

Dockerfile
niho: I think according to proper UNIX practice and the [Filesystem Hierarchy Standard](https://www.pathname.com/fhs/pub/fhs-2.3.html) you should use the `/opt` directory for "the installation of add-on application software packages". The `/usr` directory is mostly intended for stuff that is part of the distribution/OS (and `/usr/local` for stuff you build and install "manually" to not clash with the OS installed stuff). The `/usr/src` directory is most typically the home of the kernel source files on a Linux system for example.

But I guess in this context that's more in the category of "funny facts" than anything else. `/usr/src` will work just fine.

Dockerfile
```diff
@@ -1,7 +1,29 @@
...
+##### Setup Maven ####
+FROM maven:3.6.3-amazoncorretto-11 AS dependencies
 
+# Resolve dependencies and cache them
+COPY pom.xml /
+RUN mvn dependency:go-offline
+
+
+##### Build stage #####
+FROM dependencies AS build
```
niho, line 10: I don't think you gain anything by having two separate build stages for `build` and `dependencies` right?

Dockerfile
```diff
@@ -1,7 +1,29 @@
...
+
+# Fetch the datadog agent
 RUN curl -o dd-java-agent.jar -L 'https://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=com.datadoghq&a=dd-java-agent&v=LATEST'
 
-ADD target/back-office-0.0.1-SNAPSHOT.jar /
+# Copy the jar from build stage to this one
+COPY --from=build /target/back-office-0.0.1-SNAPSHOT.jar /
 
+# Define entry point
 ENTRYPOINT java -javaagent:/dd-java-agent.jar -jar back-office-0.0.1-SNAPSHOT.jar
```
niho, line 29: I think it is good practice to always include an `EXPOSE` statement in Dockerfiles, even if you typically expose the port manually when you start the container it is good documentation on what port the service lives on without having to dig through the source code.

Dockerfile
bystam: Only text output reporting 🤷 but I just merged it now :(

Dockerfile
bystam: That makes sense!

Dockerfile
niho: That's why you should never review a pull request — someone else will always approve it before you. :)

bystam: 🤔

Dockerfile
bystam: It's not obvious which ports are exposed though, it differs between services and staging/production, so I would have to dig through what goes for back-office specifically. I think it might be different on staging/production for this one actually.

Dockerfile
bystam: It's more like, since the PR had been often for some time, and I hadn't received any responses to my messages about it on Slack, I didn't expect any more reviews to show up. Then once Arechoug left one I took it as a go-ahead to merge it.

bystam: It was also a remnant from when I ran the build stage-by-stage as I described on slack, and then it made sense to split it up due to the individual progress reporting in the UI. But that is now handled by the log output itself from docker. Not as nice, but functional.

Dockerfile
bystam: Yeah I ended up changing it to just having the files at the top instead, simply because it de-cluttered the file so much. I think that will be helpful when copy-pasting this to other services.

Dockerfile
bystam: Fixed this on master now!

Dockerfile
niho: > It's not obvious which ports are exposed though

The more reason to document it in the Dockerfile then. :)

> I think it might be different on staging/production for this one actually.

It might be outside the scope of this PR, but I think a good practice is to let the executing environment (e.g. the host) handle the port mapping. So for example the production environment would say: map port `8080` to port `80`, or map port `8080` to port `5678`. And each service should then expose a predictable default port (say `8080`) that is the same for all services. Trying to assign unique hard coded ports to each service is a mess! So instead, if you need multiple services to run concurrently on the same machine/network exposed to the outside world, you would simply let Docker Compose or Kubernetes or whatever you are using do the port mapping and assign unique ports to each service (AWS ECS can allocate ports automatically and map them to a load balancer, for example). And the same goes for inter-service communication. That should not rely on fixed ports, but should be configured by the executing environment and passed in to each service using environment variables, e.g.:

```
docker run -p 8001 service-a
docker run -p 8002 -e SERVICE_A_URL=http://localhost:8001/ service-b
```

PR #342, opened 2021-02-19
Enable swish payout

FredrikAreschoug:
## What?
- Enable swish payout

## Why?
- :money_with_wings: 

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally


src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
```diff
@@ -297,8 +300,49 @@ public PaymentCompletionResponse createPaymentCompletionLink(final String member
...
+        return addPayment(id, claimPayment, env);
+    }
+
+    public CompletableFuture<DataFetcherResult<Claim>> createClaimSwishPayment(
+        UUID id,
+        ClaimSwishPaymentInput paymentInput,
+        DataFetchingEnvironment env
+    ) throws AuthorizationException {
+        log.info("Personnel with email '{}'' adding claim payment",
+            GraphQLConfiguration.getEmail(env, personnelService));
```
bystam, line 319: This places GDPR protected information (email) in our logs. Do we say explicitly that that's okay for our own employees? If not, I think it's probably something to be wary of.

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
```diff
@@ -297,8 +300,49 @@ public PaymentCompletionResponse createPaymentCompletionLink(final String member
             paymentInput.getExGratia(),
             ClaimPaymentType.valueOf(paymentInput.getType().toString()),
             GraphQLConfiguration.getEmail(env, personnelService),
-            paymentInput.isSanctionListSkipped()
+            paymentInput.isSanctionListSkipped(),
+            null,
+            null,
+            null,
+            null,
```
bystam, line 307: How come all these nulls arrived. Is that a lombok thing that it generates all the different constructors or something? And that you had to fill it with `null`s to reach the `SelectedPayoutDetails` type?

bystam (commented):
One comment and one curious question :)

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
Michael-Duivestein: Does that also count when it comes to work email accounts in an office environment?

FredrikAreschoug: We should probably remove this in more places yes

src/main/java/com/hedvig/backoffice/graphql/GraphQLMutation.java
FredrikAreschoug: Yeah Lombok makes java a bit nicer but yeah I had to add them to be able to add `SelectedPayoutDetails.NotSelected.INSTANCE`. One day we should convert this class!!

src/main/java/com/hedvig/backoffice/services/payments/PaymentServiceClient.java
```diff
@@ -36,10 +36,6 @@ void chargeMember(
     @PostMapping("/v2/_/members/{memberId}/payout")
     ResponseEntity<UUID> payoutMember(
         @PathVariable String memberId,
-        @RequestParam(name = "category") PayoutCategory category,
```
AlexSJ1990, line 39: Nice 🧹

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceStub.java
```diff
@@ -93,7 +95,8 @@ public ClaimsServiceStub() {
                 LocalDateTime.now(),
                 ClaimPaymentStatus.COMPLETED,
                 null,
-                "HDI"
+                "HDI",
+                SelectedPayoutDetails.NotSelected.INSTANCE
```
AlexSJ1990, line 99: what does the INSTANCE do?

AlexSJ1990 (approved):
✅

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceStub.java
FredrikAreschoug: In java, you need to specify that you need an instance of a kotlin object to work. To solve this kotlin create a static field called `INSTANCE`

PR #345, opened 2021-02-23
Send down carrier when making a claim payout

vonElfvin:
# Jira Issue: [] 

## What?
- Send carrier for a claim payout to claims-service

## Why?
- So we can pick correct payout account
- So we can have the correct carrier for expenses

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally


AlexSJ1990 (approved):
👍

PR #349, opened 2021-03-23
rely on the dependencyManagement declaration of spring cloud dependencies instead

bystam:
# Jira Issue: [] 

## What?
Some spring cloud kubernetes dependencies were locked to a super old version, while there was also a `dependencyManagement` declaration for them. This removes the specific old versions to instead use the somewhat more modern one declared at the top of the pom.xml file.

## Why?
We're getting lots of error logs because of bugs in the old versions.

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -195,17 +195,14 @@
     <dependency>
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-starter-kubernetes</artifactId>
-      <version>0.3.0.RELEASE</version>
     </dependency>
     <dependency>
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-kubernetes-config</artifactId>
-      <version>0.3.0.RELEASE</version>
     </dependency>
     <dependency>
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-kubernetes-ribbon</artifactId>
-      <version>0.3.0.RELEASE</version>
     </dependency>
     <dependency>
       <groupId>org.springframework.cloud</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>2.0.4.RELEASE</version>
+    <version>2.1.13.RELEASE</version>
     <relativePath/>
     <!-- lookup parent from repository -->
   </parent>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -52,7 +52,7 @@
       <dependency>
         <groupId>org.springframework.cloud</groupId>
         <artifactId>spring-cloud-dependencies</artifactId>
-        <version>Finchley.SR4</version>
+        <version>Greenwich.SR5</version>
```
Michael-Duivestein, line 55: Based on claims-service, you can theoretically upgrade to Hoxton.SR10:
https://github.com/HedvigInsurance/claims-service/pull/81/files

We don't have AWS here, and it's one of the two difficult updates.
The other one is org.javamoney.moneta.moneta-core. Updating this requires upcasters.

bystam: > We don't have AWS here, and it's one of the two difficult updates.
> The other one is org.javamoney.moneta.moneta-core. Updating this requires upcasters.

Do you mean that we don't have that in `Greenwich.SR5` or on `Hoxton.SR10`?  Which one of them requires upcasters and which one doesn't?

bystam: I am able to run and test this service with the current configuration.

pom.xml
```diff
@@ -366,7 +370,7 @@
     </dependency>
     <dependency>
       <groupId>org.jetbrains.kotlin</groupId>
-      <artifactId>kotlin-test</artifactId>
+      <artifactId>kotlin-test-junit</artifactId>
```
bystam, line 373: IntelliJ told me I should use the `-junit` one, so I did. Seems to still run the tests.

pom.xml
```diff
@@ -195,17 +200,16 @@
...
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-starter-kubernetes</artifactId>
-      <version>0.3.0.RELEASE</version>
+      <scope>runtime</scope>
     </dependency>
     <dependency>
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-kubernetes-config</artifactId>
-      <version>0.3.0.RELEASE</version>
+      <scope>runtime</scope>
```
bystam, line 208: Other services has these tagged as `runtime`. Good or bad?

pom.xml
```diff
@@ -61,6 +61,11 @@
   </dependencyManagement>
 
   <dependencies>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.2</version>
+    </dependency>
```
bystam, line 68: We used `javax.annotation.Nullable` a bit, which we seemed to have received from a transitive dependency of one of the spring ones. Bumping them made us lose that, so I added it again.

FredrikAreschoug (approved):
If this is tested I see no reason to not ship it :slightly_smiling_face:

pom.xml
vonElfvin: This might be why it is spewing out errors at startup when running locally 🤔

vonElfvin (approved):

pom.xml
bystam: How so? And what errors? The ones I wrote this PR for is the periodic errors printed every minute.

image

bystam:
> If this is tested I see no reason to not ship it 🙂

Yeah I think this is one of the things that actually *must* be tested as part of a kubernetes cluster to be sure. But there's always the revert button if it fails. :)

PR #354, opened 2021-04-15
Turn off session management and apply some code styling

palmenhq:
# Jira Issue: [] 

## What?
- It seems we created a new session per request previously, causing strange issues when sending parallell requests. As Gateekeer is responsible for authentication, and auth is the only thing we used to use sessions for (afaik), we can now remove it. This should also kill off the error message we're getting in some requests ("_This session has been expired (possibly due to multiple concurrent logins being attempted as the same user)_").

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally


meletios (approved):
Sweet!!

PR #359, opened 2021-05-21
Feat/validate member charges

fredriklagerblad:
# Jira Issue: [IPC-102] 

## What?
- Validate that the member direct charge event is equal to the current balance.

## Why?
- You should only be able to charge a member exactly the outstanding current balance.

## Optional checklist
- [x] Unit tests written
- [x] Tested locally

[IPC-102]: https://hedvig.atlassian.net/browse/IPC-102


Michael-Duivestein (approved):

AlexSJ1990 (approved):
Really great with adding tests :)

PR #360, opened 2021-05-26
Extract claim type mapping

rasmusguterstam:
# Jira Issue: [] 

## What?
- Remove claim type mapping, which from now on will happen in `claims-service`
- Kill off unused end-points
- Some good 'ol codescouting

## Why?
- Claim types and their mappings are currently spread out across `hedgehog`, `back-office` and `claims-service`. This is a first step in moving this logic to `claims-service`.
- We want to extend claim types, but we have to do this first 🤓

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally


src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceClient.java
```diff
@@ -58,9 +58,6 @@ ClaimSearchResultDTO search(
   @GetMapping("/_/claims/claim?claimID={id}")
   Claim find(@PathVariable("id") String id, @RequestHeader("Authorization") String token);
 
-  @GetMapping("/_/claims/claimTypes")
-  List<ClaimType> types(@RequestHeader("Authorization") String token);
```
vonElfvin, line 62: :waitwhat:

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceClientFallback.java
```diff
@@ -71,12 +71,6 @@ public Claim find(String id, String token) {
         return null;
     }
 
-    @Override
-    public List<ClaimType> types(String token) {
-        log.error("request to claim-service failed");
-        return new ArrayList<>();
-    }
```
vonElfvin, line 78: Why does this class even exist? Didn't know we had a "Fallback" class

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceImpl.java
```diff
@@ -145,10 +140,9 @@ public long totalClaims(String token) {
     public List<Claim> getClaimsByIds(List<UUID> ids) {
         return client.getClaimsByIds(new ClaimsByIdsDto(ids))
             .stream()
-            .map(claim -> {
+            .peek(claim -> {
```
vonElfvin, line 143: Are you sure peek is correct here? We do want the `signedAudioUrl` to be included in the result and thus `map` looks correct to me

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceStub.java
```diff
@@ -4,25 +4,8 @@
...
-import com.hedvig.backoffice.services.claims.dto.ClaimSearchResultDTO;
-import com.hedvig.backoffice.services.claims.dto.ClaimSortColumn;
-import com.hedvig.backoffice.services.claims.dto.ClaimSource;
-import com.hedvig.backoffice.services.claims.dto.ClaimStateUpdate;
-import com.hedvig.backoffice.services.claims.dto.ClaimType;
-import com.hedvig.backoffice.services.claims.dto.ClaimTypeUpdate;
-import com.hedvig.backoffice.services.claims.dto.CreateBackofficeClaimDTO;
-import com.hedvig.backoffice.services.claims.dto.EmployeeClaimRequestDTO;
-import com.hedvig.backoffice.services.claims.dto.MarkClaimFileAsDeletedDTO;
+import com.hedvig.backoffice.services.claims.dto.*;
```
vonElfvin, line 7: Oh no wildcard 😢

vonElfvin (approved):
Awesome! Feels really good with some refactoring for these things 🙏 

I'm curious how this will impact the deployment process since now we might have to deploy claim-service and back-office at the same time?

src/main/java/com/hedvig/backoffice/graphql/resolvers/ClaimResolver.kt
```diff
@@ -79,9 +79,11 @@ class ClaimResolver(
     }
 
     fun getType(claim: Claim, env: DataFetchingEnvironment?): Any? {
-        return if (claim._type == null) {
+        val claimType = claim._type
```
bystam, line 82: You can avoid the first `if` and the nested `else when` by simply chaing this line to this:
```suggestion
        val claimType = claim._type ?: return null // now claimType will not be null on the next line
```

bystam (commented):
This looks good to me, had one lite readability suggestion. Otherwise it looks like Elvin had some suggestions too :)

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceClient.java
rasmusguterstam: Yeah ikr 😩 is never used

rasmusguterstam: I guess this is a good example of how spread out the types are - someone once aimed at letting `claims-service` provide the source of truth, but the `hedgehog` types were the ones used in practice

src/main/java/com/hedvig/backoffice/services/claims/ClaimsServiceClientFallback.java
rasmusguterstam: Yeah I was surprised as well. I guess it does what it's supposed to, but every method in this fallback just logs "request to claim-service failed", and since we're not even consistent with respect to other services, I would suggest we remove this pattern and just let it fail (as with everything else)?

PR #363, opened 2021-06-23
Make pickedLocale nullable, add Duplicate and Other claim type, use ContractMarketInfo when generating payments link

vonElfvin:
# Jira Issue: [] 

## What?
- Make pickedLocale nullable
- Add Duplicate and Other claim types
- Use ContractMarketInfo when generating payments link

## Why?
- Since pickedLocale is not always present in `member-service` and we are always falling back to Swedish
- So we can go do Denmark with the `Other` claims type for Accident+Travel
- So we can mark accidently created claims as `Duplicate`
- Since the pickedLocale is not always present, we should use the insurance source as a better source of truth

## Optional checklist
- [ ] Unit tests written
- [ ] Tested locally


src/main/java/com/hedvig/backoffice/graphql/Util.java
```diff
@@ -2,125 +2,132 @@
...
+                return "Snow Pressure";
+            case StormDamageClaim:
+                return "Storm Damage";
+            case VerminAndPestsClaim:
+                return "Vermin And Pests";
+            case OtherClaim:
+                return "Other";
+            case DuplicateClaim:
+                return "Duplicate";
+            case TestClaim:
```
tobiasbexelius, line 56: Should this be in here?

tobiasbexelius: I just noticed that it's not new, but still :)

src/main/java/com/hedvig/backoffice/graphql/types/claims/DuplicateClaim.java
```diff
@@ -0,0 +1,30 @@
...
+import com.hedvig.backoffice.services.claims.dto.ClaimData;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.List;
+import lombok.Value;
+import lombok.val;
+
+@Value
+@UnionType
+public class DuplicateClaim {
```
tobiasbexelius, line 13: Convert to kotlin perhaps? Or is there a reason to create a brand new java file?

src/main/java/com/hedvig/backoffice/graphql/types/claims/OtherClaim.java
```diff
@@ -0,0 +1,56 @@
...
+import com.hedvig.backoffice.services.claims.dto.ClaimData;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.List;
+import lombok.Value;
+import lombok.val;
+
+@Value
+@UnionType
+public class OtherClaim {
```
tobiasbexelius, line 13: Same here

tobiasbexelius (commented):
Looks good, just a few questions 😄

src/main/java/com/hedvig/backoffice/graphql/Util.java
vonElfvin: Yes it should :)

src/main/java/com/hedvig/backoffice/graphql/types/claims/DuplicateClaim.java
vonElfvin: This code is deprecated as soon as it's released really. Rasmus' update to claim types will remove these files entirely. Now we just need something out to support these claim types for the summer vacation

tobiasbexelius (approved):

PR #364, opened 2021-06-24
Add trials query

tobiasbexelius:
# Jira Issue: [MX-114, MX-120] 

## What?
- 

## Why?
- 

## Optional checklist
- [ ] Unit tests written
- [x] Tested locally


src/main/java/com/hedvig/backoffice/graphql/types/Trial.kt
```diff
@@ -0,0 +1,46 @@
...
+    constructor(trialDto: TrialDto) : this(
+        id = trialDto.id,
+        memberId = trialDto.memberId,
+        fromDate = trialDto.fromDate,
+        toDate = trialDto.toDate,
+        displayName = when (trialDto.type) {
+            "SE_APARTMENT_RENT" -> "Trial insurance for rental apartment"
+            "SE_APARTMENT_BRF" -> "Trial insurance for BRF apartment"
+            else -> trialDto.type
+        },
```
vonElfvin, line 24: I personally like having dto logic like this at the source (product-pricing) so if we add a new type there we'll add the displayName and Hope will work without any more code in back-office. I do like the decoupling of reponsibilities though since it's a bit overkill for product-pricing to care about displayName 🤷

src/main/java/com/hedvig/backoffice/graphql/types/Trial.kt
```diff
@@ -0,0 +1,46 @@
+package com.hedvig.backoffice.graphql.types
+
+import com.hedvig.backoffice.services.product_pricing.dto.TrialDto
+import java.time.LocalDate
+
+data class Trial(
```
vonElfvin, line 6: Any particular reason we don't add the type (as a string) here? Could be useful for potential frontend visual logic for example

vonElfvin (approved):
Nice! :)

src/main/java/com/hedvig/backoffice/config/ExternalServicesConfig.kt
```diff
@@ -54,7 +54,7 @@ class ExternalServicesConfig @Autowired constructor(private val context: Applica
     private val rest = RestTemplateBuilder()
         // use short connect timeout to make sure we bail asap if it seems to be down,
         // saving us precious boot time
-        .setConnectTimeout(30)
+        .setConnectTimeout(1000)
```
Michael-Duivestein, line 57: Is there any specific reason why this was increased?

tobiasbexelius: Yes, 30 turned out to be too low when port-forwarding from staging 😄

src/main/java/com/hedvig/backoffice/graphql/types/Trial.kt
tobiasbexelius: Yeah, I understand your concern and like you suggest the separation of concerns was the reason for putting it here.

src/main/java/com/hedvig/backoffice/graphql/types/Trial.kt
tobiasbexelius: We don't have a need for it at the moment, so to keep it lean I would say. It's easy to add if we ever do need it

